PR #2, opened 2017-03-03
Organization Structure Project

elinaldosoft:
Organization structure tests.


elinaldosoft:
I signed it

requirements.txt
```diff
@@ -0,0 +1 @@
+ipython
```
nitzpo: Why add a requirements file for this?
It is already included in the install_requires argument for setup() in setup.py.

elinaldosoft: To configure the environment in development mode, But I believe the file name should be 
requirements_dev.txt. What do you think ?

datashaman: `python setup.py develop`

elinaldosoft: Nice, thanks

Issue #4, opened 2017-03-04
fire.Fire(func) means that `func` cannot return an empty dict


ubershmekel:
```
import fire

def x():
    return {}

fire.Fire(x)
```

Will cause this stack trace:

```
Traceback (most recent call last):
  File "E:/1stuff/Dropbox/dev/python/testfire/testing.py", line 7, in <module>
    z = fire.Fire(x)
  File "C:\Users\yuv\AppData\Local\Programs\Python\Python36-32\lib\site-packages\fire\core.py", line 145, in Fire
    _PrintResult(component_trace, verbose=component_trace.verbose)
  File "C:\Users\yuv\AppData\Local\Programs\Python\Python36-32\lib\site-packages\fire\core.py", line 181, in _PrintResult
    print(_DictAsString(result, verbose))
  File "C:\Users\yuv\AppData\Local\Programs\Python\Python36-32\lib\site-packages\fire\core.py", line 201, in _DictAsString
    len(str(key)) for key in result.keys()
ValueError: max() arg is an empty sequence
```

guoruibiao:
python-fire is awesome cool.

dbieber:
@ubershmekel Thanks for finding and fixing this bug! I'll respond to the pull request later in the week.

PR #5, opened 2017-03-04
Fix for empty dictionary print outs

ubershmekel:
This should handle https://github.com/google/python-fire/issues/4

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -198,7 +198,7 @@ def _DictAsString(result, verbose=False):
     A string representing the dict
   """
   if not result:
-    return "{}"
+    return '{}'
   
   longest_key = max(
       len(str(key)) for key in result.keys()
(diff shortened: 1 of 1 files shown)


ubershmekel:
I signed it!

dbieber:
Thanks for the fix! Looks good! One small change to make: let's use single quotes to maintain consistency in the project.

----------

I also want to call out that I agree with your decision to return '{}' rather than the empty string (which is what would have been returned had the existing code been working). I prefer '{}' so the user can tell that there is a result and that it is a dict.

Relatedly, in the future we may want to make additional changes to how results are printed. As an example, it may make sense to display an empty list as '[]' (currently it will display nothing for an empty list). At that point we'll want to do a full review of our serialization methods, rather than making these changes one by one.

Issue #6, opened 2017-03-05
annotations and kwonlyargs


jgarst:
inspectutils.GetArgSpec uses inspect.getargspec, which can't handle annotations and kwonly args. The program

```
from fire import Fire

def test_args(arg1: str = 'aarrrgh', *, arg2: str = 'arrrgh'):
    return arg2, arg2

if __name__ == '__main__':
    Fire(test_args)
```
leads to an error at line 75 in GetArgSpec
"ValueError: Function has keyword-only parameters or annotations, use getfullargspec() API which can support them"

The simplest fix would be to put in another except clause to catch the value error, and proceed with inspect.getfullargspec or inspect.signature instead. An alternate way might be to decide at import if the program is python 3 and inspect.signature can be called. Do you have a preference between the two solutions, or another suggestion?

-Jared

dbieber:
Thanks for finding and reporting this bug. I expect that checking the python version with six and then using inspect.getfullargspec for Python 3 will be the appropriate way to proceed.

jgarst:
I've just submitted a pull request for the annotations part of this issue. The keyword only argument part of this bug bleeds over to a more invasive enhancement request, as you probably want to maintain the keyword only property on the command line.

It also seems like you are doing a lot of work to reproduce [inspect.signature](https://www.python.org/dev/peps/pep-0362/), such as dropping self and loading the arguments tuple by hand. You might want to look at a [backport](https://github.com/aliles/funcsigs), though don't know the code quality, or your stance on dependencies.

hapnewyear2017:
from fire import Fire

def test_args(arg1: str = 'aarrrgh', *, arg2: str = 'arrrgh'):
    return arg2, arg2

if __name__ == '__main__':
    Fire(test_args)

Cologler:
Is possible auto convert argument type?

like:

``` py
def func(arg: int):
    print(type(arg)) # print int
```

dbieber:
Merged #18

PR #8, opened 2017-03-05
Fix error with annotated functions.

jgarst:
Pull request for bug in issue #6. Note that it does not fix the stickier issue with keyword only arguments.

### DIFF
--- a/fire/inspectutils.py
+++ b/fire/inspectutils.py
@@ -76,7 +76,7 @@ def GetArgSpec(fn):
     if six.PY2:
       argspec = inspect.getargspec(fn)
       keywords = argspec.keywords
-    elif six.PY3:
+    else:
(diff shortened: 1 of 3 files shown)


jgarst:
I signed it!

fire/inspectutils.py
```diff
@@ -72,11 +72,18 @@ def GetArgSpec(fn):
   fn, skip_arg = _GetArgSpecFnInfo(fn)
 
   try:
-    argspec = inspect.getargspec(fn)
+
+    if six.PY2:
+      argspec = inspect.getargspec(fn)
+      keywords = argspec.keywords
+    elif six.PY3:
```
dbieber: nit: Let's change this to "else:" for consistency with other PY2 checks, and so that a reader/static analyzer can guarantee that one of the two branches will be taken.

PR #9, opened 2017-03-05
cleanup logic in print

r0fls:
This logic can be condensed.

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -57,6 +57,7 @@ import json
 import pipes
 import shlex
 import sys
+import types
 
 from fire import completion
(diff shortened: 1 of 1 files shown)


r0fls:
I signed it!

dbieber:
Thanks for this clean up.
Let's keep the tuple clause separate from the others (at least for now) so we don't change any behavior with this clean up. The reason tuples are printed on one line is that they are commonly meant to be read as a tuple (e.g. for coordinates), rather than as a list of items.

It may make sense to change/unify this behavior in the future, but that will require a longer discussion first.

PR #18, opened 2017-03-11
Fix keyword-only arguments

jgarst:
Should fix the rest of #6.

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -546,16 +546,19 @@ def _MakeParseFn(fn):
     can then be called with fn(*varargs, **kwargs). The remaining_args are
     the leftover args from the arguments to the parse function.
   """
-  fn_args, fn_varargs, fn_keywords, fn_defaults = inspectutils.GetArgSpec(fn)
+  (fn_args, fn_varargs, fn_keywords, fn_defaults, fn_kwonly,
+    fn_kwonlydefaults, fn_annotations) = inspectutils.GetArgSpec(fn)
   metadata = decorators.GetMetadata(fn)
 
-  # Note: num_required_args is the number of arguments without default values.
-  # All of these arguments are required.
+  # Note: num_required_args is the number of positional arguments
+  # without default values. All of these arguments are required.
   num_required_args = len(fn_args) - len(fn_defaults)
+  required_kwonly = set(fn_kwonly) - set(fn_kwonlydefaults)
 
   def _ParseFn(args):
     """Parses the list of `args` into (varargs, kwargs), remaining_args."""
-    kwargs, remaining_args = _ParseKeywordArgs(args, fn_args, fn_keywords)
+    kwargs, remaining_args = _ParseKeywordArgs(args, fn_args + fn_kwonly,
+                                               fn_keywords)
 
     # Note: _ParseArgs modifies kwargs.
     parsed_args, kwargs, remaining_args, capacity = _ParseArgs(
@@ -566,8 +569,13 @@ def _MakeParseFn(fn):
       # If we're allowed *varargs or **kwargs, there's always capacity.
       capacity = True
 
-    if fn_keywords is None and kwargs:
-      raise FireError('Unexpected kwargs present', kwargs)
+    extra_kw = set(kwargs) - set(fn_kwonly)
+    if fn_keywords is None and extra_kw:
+      raise FireError('Unexpected kwargs present:', extra_kw)
+
+    missing_kwonly = set(required_kwonly) - set(kwargs)
+    if missing_kwonly:
+      raise FireError('Missing required flags:', missing_kwonly)
 
     # If we accept *varargs, then use all remaining arguments for *varargs.
     if fn_varargs is not None:
@@ -580,7 +588,7 @@ def _MakeParseFn(fn):
 
     varargs = parsed_args + varargs
 
-    consumed_args = args[:len(args)-len(remaining_args)]
+    consumed_args = args[:len(args) - len(remaining_args)]
     return (varargs, kwargs), consumed_args, remaining_args, capacity
 
   return _ParseFn
--- a/fire/core.py
+++ b/fire/core.py
@@ -546,12 +546,12 @@ def _MakeParseFn(fn):
     can then be called with fn(*varargs, **kwargs). The remaining_args are
     the leftover args from the arguments to the parse function.
   """
-  fn_spec = inspectutils.specification(fn)
+  fn_spec = inspectutils.GetFullArgSpec(fn)
   all_args = fn_spec.args + fn_spec.kwonlyargs
   metadata = decorators.GetMetadata(fn)
 
-  # Note: num_required_args is the number of positional arguments
-  # without default values. All of these arguments are required.
+  # Note: num_required_args is the number of positional arguments without
+  # default values. All of these arguments are required.
   num_required_args = len(fn_spec.args) - len(fn_spec.defaults)
   required_kwonly = set(fn_spec.kwonlyargs) - set(fn_spec.kwonlydefaults)
 
--- a/fire/core.py
+++ b/fire/core.py
@@ -546,31 +546,30 @@ def _MakeParseFn(fn):
     can then be called with fn(*varargs, **kwargs). The remaining_args are
     the leftover args from the arguments to the parse function.
   """
-  (fn_args, fn_varargs, fn_keywords, fn_defaults, fn_kwonly,
-    fn_kwonlydefaults, fn_annotations) = inspectutils.GetArgSpec(fn)
+  fn_spec = inspectutils.specification(fn)
+  all_args = fn_spec.args + fn_spec.kwonlyargs
   metadata = decorators.GetMetadata(fn)
 
   # Note: num_required_args is the number of positional arguments
   # without default values. All of these arguments are required.
-  num_required_args = len(fn_args) - len(fn_defaults)
-  required_kwonly = set(fn_kwonly) - set(fn_kwonlydefaults)
+  num_required_args = len(fn_spec.args) - len(fn_spec.defaults)
+  required_kwonly = set(fn_spec.kwonlyargs) - set(fn_spec.kwonlydefaults)
 
   def _ParseFn(args):
     """Parses the list of `args` into (varargs, kwargs), remaining_args."""
-    kwargs, remaining_args = _ParseKeywordArgs(args, fn_args + fn_kwonly,
-                                               fn_keywords)
+    kwargs, remaining_args = _ParseKeywordArgs(args, all_args, fn_spec.varkw)
 
     # Note: _ParseArgs modifies kwargs.
     parsed_args, kwargs, remaining_args, capacity = _ParseArgs(
-        fn_args, fn_defaults, num_required_args, kwargs, remaining_args,
+        fn_spec.args, fn_spec.defaults, num_required_args, kwargs, remaining_args,
         metadata)
 
-    if fn_varargs or fn_keywords:
+    if fn_spec.varargs or fn_spec.varkw:
       # If we're allowed *varargs or **kwargs, there's always capacity.
       capacity = True
 
-    extra_kw = set(kwargs) - set(fn_kwonly)
-    if fn_keywords is None and extra_kw:
+    extra_kw = set(kwargs) - set(fn_spec.kwonlyargs)
+    if fn_spec.varkw is None and extra_kw:
       raise FireError('Unexpected kwargs present:', extra_kw)
 
     missing_kwonly = set(required_kwonly) - set(kwargs)
@@ -578,7 +577,7 @@ def _MakeParseFn(fn):
       raise FireError('Missing required flags:', missing_kwonly)
 
     # If we accept *varargs, then use all remaining arguments for *varargs.
-    if fn_varargs is not None:
+    if fn_spec.varargs is not None:
       varargs, remaining_args = remaining_args, []
     else:
       varargs = []
--- a/fire/fire_test.py
+++ b/fire/fire_test.py
@@ -20,8 +20,9 @@ import fire
 from fire import test_components as tc
 from fire import trace
 
-import unittest
 import six
+import unittest
+
 
 class FireTest(unittest.TestCase):
 
@@ -102,7 +103,7 @@ class FireTest(unittest.TestCase):
     self.assertEqual(fire.Fire(tc.Annotations, 'double 5'), 10)
     self.assertEqual(fire.Fire(tc.Annotations, 'triple 5'), 15)
 
-  @unittest.skipIf(six.PY2, 'keyword only arguments not supported in python 2')
+  @unittest.skipIf(six.PY2, 'Keyword-only arguments not supported in Python 2')
   def testFireKeywordOnlyArgs(self):
     self.assertIsNone(fire.Fire(tc.py3.KeywordOnly, 'double 5'))
 
--- a/fire/fire_test.py
+++ b/fire/fire_test.py
@@ -20,9 +20,8 @@ import fire
 from fire import test_components as tc
 from fire import trace
 
-import sys
 import unittest
-
+import six
 
 class FireTest(unittest.TestCase):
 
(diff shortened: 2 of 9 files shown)


dbieber:
Awesome! I'll look through the pull request later in the week.

We'll also need to get travis passing before we can check this in, and there'll be some style things to fix up too. My first thought for how to make Travis happy is to conditionally exclude files from nosetest (called in [.travis.yml](https://github.com/google/python-fire/blob/master/.travis.yml) atm) based on the python version, which I think Travis exposes as an env variable. Open to suggestions.

jgarst:
Ach, I missed that you included a travis file, and tried to test by hand. Sorry about that.

That nose test discovery is annoyingly broad. I suppose I like starting a nose exclude list, if the alternative is removing the word 'test' from test related non-scripts. I've updated the nose command in travis.yml. An alternative might be to create a separate test directory. Then testing components would not need 'test' in their name for clarity, and so would not be discovered and run automatically. This is getting way into the minutia though.

Let me know about the style issues. I tried to be conservative and match yours, but I will have of course missed some bits.

The non-CI related problem travis discovered is a bug in python < 3.4.3 namedtuple._asdict. Specifically subclasses of namedtuple always returning an empty dictionary. 

The workaround isn't that ugly, but having a datatype with that much of a gotcha is a landmine for development. Unless you are looking for an excuse to drop support for python 3.4.0 - 3.4.3, I'll figure out some way to avoid it.

fire/inspectutils.py
```diff
@@ -66,43 +87,37 @@ def GetArgSpec(fn):
     A named tuple of type inspect.ArgSpec with the following fields:
```
dbieber: It's no longer of type inspect.ArgSpec.

fire/inspectutils.py
```diff
@@ -66,43 +87,37 @@ def GetArgSpec(fn):
...
       args: A list of the argument names accepted by the function.
       varargs: The name of the *varargs argument or None if there isn't one.
-      keywords: The name of the **kwargs argument or None if there isn't one.
+      varkw: The name of the **kwargs argument or None if there isn't one.
       defaults: A tuple of the defaults for the arguments that accept defaults.
+      kwonlyargs: A list of argument names that must be passed with a keyword.
```
dbieber: We'll have to update completion.py as well:
https://github.com/google/python-fire/blob/master/fire/completion.py#L161

fire/helputils.py
```diff
@@ -126,7 +126,8 @@ def HelpString(component, trace=None, verbose=False):
   return '\n'.join(lines)
 
 
-def _UsageStringFromFnDetails(command, args, varargs, keywords, defaults):
+def _UsageStringFromFnDetails(command, args, varargs, varkw, defaults,
```
dbieber: Let's just have this function accept an argspec instead of a long list of parameters.
And as long as we're here let's rename the function so that we have consistent terminology.

"def _UsageStringFromArgSpec(command, argspec):"

(This function is only used in one place, so no need to go searching for references.)

(I'm actually inclined to rename all our uses of the word "argspec" to "signature", but we can do that separately from this fix.)

fire/inspectutils.py
```diff
@@ -22,6 +22,27 @@
...
+    kwonlyargs=None, kwonlydefaults=None, annotations=None):
+    return super(_FullArgSpec, cls).__new__(
+      cls,
+      args or [],
+      varargs,
+      varkw,
+      defaults or (),
+      kwonlyargs or [],
+      kwonlydefaults or {},
+      annotations or {}
```
dbieber: trailing comma
and two extra spaces on all these lines (38-44)

fire/inspectutils.py
```diff
@@ -22,6 +22,27 @@
 
 import IPython
 import six
+from collections import namedtuple
```
dbieber: move above "import inspect"

fire/inspectutils.py
```diff
@@ -22,6 +22,27 @@
...
+
+_FullArgTuple = namedtuple('FullArgSpec', ('args', 'varargs', 'varkw',
+  'defaults', 'kwonlyargs', 'kwonlydefaults', 'annotations'))
```
dbieber: format this as:

```python
_FullArgTuple = namedtuple(
    'FullArgSpec',
    ('args', 'varargs', 'varkw', 'defaults',
     'kwonlyargs', 'kwonlydefaults', 'annotations'))
```

fire/inspectutils.py
```diff
@@ -22,6 +22,27 @@
...
+
+
+class _FullArgSpec(_FullArgTuple):
+  """NamedTuple storing arguments of a callable, populated with defaults."""
+
+  def __new__(cls, args=None, varargs=None, varkw=None, defaults=None,
+    kwonlyargs=None, kwonlydefaults=None, annotations=None):
```
dbieber: align kwonlyargs with cls.

fire/inspectutils.py
```diff
@@ -22,6 +22,27 @@
 
 import IPython
 import six
+from collections import namedtuple
+
+_FullArgTuple = namedtuple('FullArgSpec', ('args', 'varargs', 'varkw',
+  'defaults', 'kwonlyargs', 'kwonlydefaults', 'annotations'))
+
+
+class _FullArgSpec(_FullArgTuple):
```
dbieber: Thinking aloud:
Is a class that inherits from a namedtuple really the best choice here?
My initial aversion is that it feels clunky. Also, now that we have 7 args, it strikes me that variable expansion is prone to breaking if we mess up the argument order somewhere and don't notice.
My preference would be to abandon the tuple aspect. And instead I think we should just have a class, and always access the members by name. Something like:

```
class FullArgSpec(object):
  """The full argspec of a callable."""

  def __init__(self, args=None, varargs=None, varkw=None, defaults=None,
               kwonlyargs=None, kwonlydefaults=None, annotations=None):
    self.args = args or []
    self.varargs = varargs
    self.varkw = varkw
    self.defaults = defaults or ()
    self.kwonlyargs = kwonlyargs or []
    self.kwonlydefaults = kwonlydefaults or {}
    self.annotations = annotations or {}
```

How's that sound?

fire/helputils.py
```diff
@@ -178,8 +190,9 @@ def UsageString(component, trace=None, verbose=False):
   command = trace.GetCommand() + ' ' if trace else ''
 
   if inspect.isroutine(component) or inspect.isclass(component):
-    args, varargs, keywords, defaults = inspectutils.GetArgSpec(component)
-    return _UsageStringFromFnDetails(command, args, varargs, keywords, defaults)
+    spec = inspectutils.GetArgSpec(component)
+    spec_dict = dict(zip(spec._fields, iter(spec)))
+    return _UsageStringFromFnDetails(command, **spec_dict)
```
dbieber: in line w/ the previous comment, this would become _UsageStringFromFnDetails(command, spec).

fire/test_components.py
```diff
@@ -17,10 +17,20 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+import six
```
dbieber: new line before import

dbieber:
Also make sure you read [this comment](https://github.com/google/python-fire/pull/18#discussion_r105549303) first, because a number of the style comments become irrelevant after switching from namedtuple to class.

fire/core.py
```diff
@@ -566,8 +569,13 @@ def _ParseFn(args):
...
 
-    if fn_keywords is None and kwargs:
-      raise FireError('Unexpected kwargs present', kwargs)
+    extra_kw = set(kwargs) - set(fn_kwonly)
+    if fn_keywords is None and extra_kw:
+      raise FireError('Unexpected kwargs present:', extra_kw)
+
+    missing_kwonly = set(required_kwonly) - set(kwargs)
+    if missing_kwonly:
+      raise FireError('Missing required flags:', missing_kwonly)
```
jgarst: Do you plan to keep this error printing to the terminal when a user mistypes something on the command line? Someone using the final command line program might be familiar with command line flags, but not the python specific concept of keyword only arguments.

fire/fire_test.py
```diff
@@ -102,6 +103,14 @@ def testFireAnnotatedArgs(self):
     self.assertEqual(fire.Fire(tc.Annotations, 'double 5'), 10)
     self.assertEqual(fire.Fire(tc.Annotations, 'triple 5'), 15)
 
+  @unittest.skipIf(sys.version_info < (3, 0),
```
jgarst: You are correct, and I have fixed it, but these testing libraries tend to only get more version specific over time. You might end up getting back to sys.version_info anyway.

fire/test_components_py3.py
```diff
@@ -0,0 +1,30 @@
...
+"""Thie module has components that use python 3 specific syntax."""
+
+
+def identity(arg1, arg2: int, arg3=10, arg4: int = 20, *arg5,
+  arg6, arg7: int, arg8=30, arg9: int = 40, **arg10):
+
+  return arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10
+
+
+class KeywordOnly:
```
jgarst: Fixed, but since this file can't be parsed by python 2 anyway, does it matter?

fire/test_components.py
```diff
@@ -17,10 +17,20 @@
...
-  return arg1, arg2, arg3, arg4
+
+def identity(arg1, arg2, arg3=10, arg4=20, *arg5, **arg10):
+  return arg1, arg2, arg3, arg4, arg5, arg10
+
+identity.__annotations__ = {'arg2': int, 'arg4': int}
+
+
+if six.PY3:
+  from .test_components_py3 import identity  # pylint: disable=unused-import
```
jgarst: That is much better, thanks.

fire/core.py
dbieber: That's a fair point.
Yeah, currently this error does surface to the user, so your string sgtm.
There's definitely room to clean up the Fire trace output, so we can think about changing how these errors are displayed if/when we go down that route.

fire/inspectutils.py
jgarst: I agree, the tuple is a holdover from python3's inspect.getargspec that I did not reevaluate. 

As this object adds more information, it might be wise to converge it to python 3's inspect.Signature object, which seems to be the [canonical future](https://www.python.org/dev/peps/pep-0362/) of signatures in python. 

Signature has an ordered dict of Parameter (name, kind, default, annotation) objects. I think this might make cleaner parsing in core._ParseFn, and I worry that without this representation your signature and python3's signature will evolve in different directions. It is also probably overpowered for what you need, and an example of small pull requests turning into larger pull requests.

The right answer probably depends on your ideas about the scope of this project. I'm already mucking about here, and happy to modify it as you like.

fire/inspectutils.py
jgarst: Fixed the line you linked here. It would probably do to bang on the completion script a little, since the args done changed.

fire/helputils.py
jgarst: For the name of the variable, I think there is virtue in trying to adhere to the conventions of the inspect module, which keeps argspec as you currently use it, and signature as the ordered dictionary of arguments I described above.

jgarst:
Thanks for taking the time to review this, by the way.

Quick question, after making the spec class a simple object, the test testGetArgSpecReturnType becomes. 

`  def testGetArgSpecReturnType(self):
    # Asserts that the object returned by GetArgSpec has the appropriate fields.
    spec = inspectutils.GetArgSpec(tc.identity)

    self.assertTrue(hasattr(spec, 'args'))
    self.assertTrue(hasattr(spec, 'defaults'))
    self.assertTrue(hasattr(spec, 'varargs'))
    self.assertTrue(hasattr(spec, 'varkw'))
    self.assertTrue(hasattr(spec, 'kwonlyargs'))
    self.assertTrue(hasattr(spec, 'kwonlydefaults'))
    self.assertTrue(hasattr(spec, 'annotations'))`

It is not totally clear what this is testing anymore, that other tests aren't doing implicitly. My inclination would be to remove it, but I didn't want to do that without comment. Let me know what you think.

dbieber:
Yes, I think it's good to remove that `testGetArgSpecReturnType` test now.

fire/inspectutils.py
```diff
@@ -63,46 +77,40 @@ def GetArgSpec(fn):
   Args:
     fn: The function to analyze.
   Returns:
-    A named tuple of type inspect.ArgSpec with the following fields:
+    An object containing the following fields:
       args: A list of the argument names accepted by the function.
```
dbieber: Let's move these arg descriptions to the docstring of _FullArgSpec's init, and then the docstring here can be much simpler, something like:

```
...
Returns:
  A FullArgSpec object for the provided function.
```

fire/inspectutils.py
```diff
@@ -24,6 +24,20 @@
 import six
 
 
+class _FullArgSpec(object):
```
dbieber: This is used outside of this module, albeit never reference by name outside of this module. So, I think we should drop the underscore. What do you think?

fire/inspectutils.py
```diff
@@ -63,46 +77,40 @@ def GetArgSpec(fn):
...
   fn, skip_arg = _GetArgSpecFnInfo(fn)
 
   try:
 
     if six.PY2:
       argspec = inspect.getargspec(fn)
-      keywords = argspec.keywords
+      args = argspec.args[1:] if skip_arg else argspec.args
+      annotations = getattr(fn, '__annotations__', None)
+      fullspec = _FullArgSpec(args, *argspec[1:], annotations=annotations)
```
dbieber: This makes me a little nervous.

Brainstorming why:
1: It hides which arguments are being passed from getargspec to _FullArgSpec,
2: subject to breaking if either getargspec or _FullArgSpec change (not really a concern I guess)
3: Reduces the approachability of the codebase (as a side effect of 1, plus using slices and *args in the same fn call).

None of these are big issues and this is nicely compact, so I guess it's fine to leave. Up to you.

Applies to line 101 as well.

fire/inspectutils.py
```diff
@@ -63,46 +77,40 @@ def GetArgSpec(fn):
...
+      args = argspec.args[1:] if skip_arg else argspec.args
+      fullspec = _FullArgSpec(args, *argspec[1:])
 
   except TypeError:
-    args = []
-    defaults = ()
     # If we can't get the argspec, how do we know if the fn should take args?
     # 1. If it's a builtin, it can take args.
     # 2. If it's an implicit __init__ function (a 'slot wrapper'), take no args.
     # Are there other cases?
```
dbieber: [Unrelated comment]
Heads up that yes, there are other cases, so we'll have to change this logic a little bit (not as part of this review, of course). This does not affect your change and will go in after.
The cases we originally missed are C-extensions, which inspect.signature/inspect.fullargspec cannot get the argspec for. While we can't solve this in the general case, we can add known argspecs for builtin modules. The only ones on my list to add so far are object.\_\_init\_\_ (which gets used a lot and which we already handle correctly) and type.\_\_init\_\_ (which is much less used, but still important).

fire/inspectutils_test.py
```diff
@@ -16,61 +16,84 @@
...
+    self.assertEqual(spec.args, ['arg1', 'arg2', 'arg3', 'arg4'])
+    self.assertEqual(spec.defaults, (10, 20))
+    self.assertEqual(spec.varargs, 'arg5')
+    self.assertEqual(spec.varkw, 'arg6')
+    self.assertEqual(spec.kwonlyargs, [])
+    self.assertEqual(spec.kwonlydefaults, {})
+    self.assertEqual(spec.annotations, {'arg2': int, 'arg4': int})
+
+
+  @unittest.skipIf(six.PY2, "No keyword arguments in python 2")
```
dbieber: single quotes

fire/fire_test.py
```diff
@@ -102,6 +102,13 @@ def testFireAnnotatedArgs(self):
     self.assertEqual(fire.Fire(tc.Annotations, 'double 5'), 10)
     self.assertEqual(fire.Fire(tc.Annotations, 'triple 5'), 15)
 
+  @unittest.skipIf(six.PY2, "keyword only arguments not supported in python 2")
```
dbieber: single quotes

fire/inspectutils.py
jgarst: I think you are right, my hesitation is that once it is a public non-builtin type it should be able to instantiate itself from a function, and other things people will reasonably expect of it.

I suppose this isn't really a reason not to do it, but it will be trickier to decide what it's job is once it is publically visible.

fire/inspectutils.py
jgarst: Oops. Missed this pretty bad.

fire/inspectutils.py
jgarst: It is probably too clever. I'll stop inling the if and just pop off the first argument. I'll probably change FullArgSpec so that it takes a function. Then passing hidden arguments won't be a problem. Since FullArgSpec no longer really behaves like inspect.FullArgSpec, it probably also needs a new name.
``

fire/inspectutils.py
jgarst: Yeah, callable objects too, like functools.partial. There is a 50/50 chance I am missing something, but I think those are basically all the cases.

You will probably be special casing forever of course. The first one should probably be special logic to grab the real arguments from the partial object, instead of the generic *args, **kwargs it provides.  My assumption is that you will eventually want something that is or looks like single dispatch here.

dbieber:
I hope you don't mind I pushed some changes directly to the pull request rather than leaving comments. See them here 157200ef7957799c61748a673d34a234234fa77d.

dbieber:
I'll wait for you to look through those changes before I move anything into google/python-fire:master. Let me know if you have comments.

Re: the naming changes:
While it probably would have been better to go with lower case function names throughout the project, I don't want to partially switch over. I want to keep consistency within the project.

jgarst:
Haha, no problem. I'm just sorry the long tail of this has taken so much of your time.

In terms of the rename, I wasn't trying to change the naming convention. I was more worried about the conceptual break that has happened with inspect.FullArgSpec, and then the problem of having the constructor for a Specification from a function is called GetArgSpecs. Someone who knew they needed a Specification object might not immediately think to call that.

Maybe this worry is overblown though, and you are the only one with a good sense for the cost of refactoring things.

PR #21, opened 2017-03-15
Fix typing error

tux-00:


dbieber:
Thanks for fixing this typo. In order merge this commit in, we need to appease the googlebot :P. If you've already signed a CLA, can you say something like "I signed it!" to make the googlebot happy?

tux-00:
I signed it!

PR #23, opened 2017-03-16
PEP8 changes

rahulkp220:
Improved the Pylint score.

### DIFF
--- a/fire/trace.py
+++ b/fire/trace.py
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""This module has classes for tracing the execution of a Fire execution.
+"""
+This module has classes for tracing the execution of a Fire execution.
 
 A FireTrace consists of a sequence of FireTraceElement objects. Each element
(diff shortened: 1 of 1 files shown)


rahulkp220:
I signed it!

waldyrious:
For convenience, note that this diff is much easier to review in ignore-whitespace mode: https://github.com/google/python-fire/pull/23/files?diff=split&w=1

dbieber:
Thanks for your contribution. While I'm open to style changes, it's important that we remain stylistically consistent across the project.

For now we're following the internal Google Python style guide. The most notable difference from the [external one](https://google.github.io/styleguide/pyguide.html) is that it permits capital-letter FunctionNames. While I would like to switch to a public style guide, it's important that we maintain consistency within the project and also that we maintain the bare-minimum level of backwards compatibility (i.e. exposing a function named `Fire`).

Let's hold off on any widespread style changes until I figure out how to release the pylintrc that we're using internally.

rahulkp220:
Ohkay sure. I agree with you on the point of consistency across the project.

PR #27, opened 2017-03-22
Exit with a status code if Fire encounters an error

dbieber:
Exit with a status code if Fire encounters an error
Introduces the "FireExit" exception, a subclass of SystemExit.

PR #28, opened 2017-03-22
FireExit: ape argparse, flesh out test cases and get everything to pass

jtratner:
@dbieber 

1. Update documentation slightly (I also changed `__init__` name to be `code` because that matches the actual property name on SystemExit)
2. Get all tests to pass and note some additional edge cases (specifically, I'm suggesting we differentiate exit codes from FireError (`2`) from exit codes from user exceptions (`1`))
3. add helper method to capture all stdout and make it easier to write regexps for comparison (i.e. check that we show "the proper..." message on bad use of help)
4. on proper `-- --help` do the same thing as argparse and exit 0

For 4:

```
# given example.py
try:
    argparse.ArgumentParser().parse_args()
except BaseException as exc:
    print type(exc), exc
    raise


›› python example.py blah; echo "Exit status $?"
usage: example.py [-h]
example.py: error: unrecognized arguments: blah
SystemExit 2
Exit status 2
›› python example.py --help; echo "Exit status $?"                                                                                                                         usage: example.py [-h]

optional arguments:
  -h, --help  show this help message and exit
SystemExit 0
Exit status 0
```

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -92,9 +92,14 @@ def Fire(component=None, command=None, name=None):
     it's a class). When all arguments are consumed and there's no function left
     to call or class left to instantiate, the resulting current component is
     the final result.
-    If the trace command line argument is supplied, the FireTrace is returned.
+    If a Fire error is encountered, the Fire Trace is displayed to stdout and
+    a FireExit is raised.
+    If the trace command line argument is supplied, the FireTrace is returned
+    and no FireExit is raised.
   Raises:
-    FireExit: If a FireError is countered.
+    FireExit: If a FireError is encountered and trace command line argument is
+        not passed. FireExit will have `code` of 2, whereas code exceptions
+        will be bubbled up as exit status of 1.
   """
   # Get args as a list.
   if command is None:
@@ -127,7 +132,7 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    raise FireExit(1, component_trace)
+    raise FireExit(2, trace)
   elif component_trace.show_trace and component_trace.show_help:
     print('Fire trace:\n{trace}\n'.format(trace=component_trace))
     result = component_trace.GetResult()
@@ -141,7 +146,8 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    return None
+    # ape argparse by exiting out quickly
+    raise FireExit(0, None)
   else:
     _PrintResult(component_trace, verbose=component_trace.verbose)
     result = component_trace.GetResult()
@@ -166,13 +172,16 @@ class FireExit(SystemExit):
 
   Contains the trace of the Fire program.
 
-  If not caught, then a FireExit will cause the program to exit with a non-zero
-  status code.
+  Attributes:
+    code (int): exit code for CLI
+    trace (FireTrace): the component trace generated from running the command.
+  Note:
+    Since this exception inherits from SystemExit/BaseException, you'll have to
+    explicitly catch it with `except SystemExit` or `except FireExit`.
   """
-
-  def __init__(self, status, data):
-    super(FireExit, self).__init__(status)
-    self.trace = data
+  def __init__(self, code, trace):
+    super(FireExit, self).__init__(code)
+    self.trace = trace
 
 
 def _PrintResult(component_trace, verbose=False):
--- a/fire/core.py
+++ b/fire/core.py
@@ -170,20 +170,20 @@ class FireError(Exception):
 class FireExit(SystemExit):
   """An exception raised by Fire to the client in the case of a FireError.
 
-  Contains the trace of the Fire program.
+  The trace of the Fire program is available on the `trace` property.
 
   Note:
     Since this exception inherits from SystemExit/BaseException, you'll have to
     explicitly catch it with `except SystemExit` or `except FireExit`.
   """
-  def __init__(self, code, trace):
+  def __init__(self, code, component_trace):
     """
     Args:
       code (int): exit code for CLI
-      trace (FireTrace): the component trace generated from running the command.
+      component_trace (FireTrace): the component trace generated from running the command.
     """
     super(FireExit, self).__init__(code)
-    self.trace = trace
+    self.trace = component_trace
 
 
 def _PrintResult(component_trace, verbose=False):
--- a/fire/core.py
+++ b/fire/core.py
@@ -132,7 +132,7 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    raise FireExit(2, trace)
+    raise FireExit(2, component_trace)
   elif component_trace.show_trace and component_trace.show_help:
     print('Fire trace:\n{trace}\n'.format(trace=component_trace))
     result = component_trace.GetResult()
@@ -147,7 +147,7 @@ def Fire(component=None, command=None, name=None):
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
     # ape argparse by exiting out quickly
-    raise FireExit(0, None)
+    raise FireExit(0, component_trace)
   else:
     _PrintResult(component_trace, verbose=component_trace.verbose)
     result = component_trace.GetResult()
--- a/fire/core.py
+++ b/fire/core.py
@@ -97,9 +97,8 @@ def Fire(component=None, command=None, name=None):
     If the trace command line argument is supplied, the FireTrace is returned
     and no FireExit is raised.
   Raises:
-    FireExit: If a FireError is encountered and trace command line argument is
-        not passed. FireExit will have `code` of 2, whereas code exceptions
-        will be bubbled up as exit status of 1.
+    FireExit: If a FireError (code 2) or exception in client code (code 1) is
+        encountered.
   """
   # Get args as a list.
   if command is None:
@@ -172,14 +171,16 @@ class FireExit(SystemExit):
 
   Contains the trace of the Fire program.
 
-  Attributes:
-    code (int): exit code for CLI
-    trace (FireTrace): the component trace generated from running the command.
   Note:
     Since this exception inherits from SystemExit/BaseException, you'll have to
     explicitly catch it with `except SystemExit` or `except FireExit`.
   """
   def __init__(self, code, trace):
+    """
+    Args:
+      code (int): exit code for CLI
+      trace (FireTrace): the component trace generated from running the command.
+    """
     super(FireExit, self).__init__(code)
     self.trace = trace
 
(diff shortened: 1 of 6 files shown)


jtratner:
@dbieber - will merge into #27

fire/core.py
```diff
@@ -92,9 +92,14 @@ def Fire(component=None, command=None, name=None):
     it's a class). When all arguments are consumed and there's no function left
     to call or class left to instantiate, the resulting current component is
     the final result.
-    If the trace command line argument is supplied, the FireTrace is returned.
+    If a Fire error is encountered, the Fire Trace is displayed to stdout and
+    a FireExit is raised.
+    If the trace command line argument is supplied, the FireTrace is returned
```
jtratner: @dbieber - If you pass `--trace` and also have invalid params, should it exit 0 or non-zero, I wasn't sure?

fire/core.py
```diff
@@ -166,13 +172,15 @@ class FireExit(SystemExit):
...
+  Note:
+    Since this exception inherits from SystemExit/BaseException, you'll have to
+    explicitly catch it with `except SystemExit` or `except FireExit`.
   """
-
-  def __init__(self, status, data):
-    super(FireExit, self).__init__(status)
-    self.trace = data
+  def __init__(self, code, trace):
+    super(FireExit, self).__init__(code)
```
jtratner: just changed name to harmonize with SystemExit property name

fire/core_test.py
```diff
@@ -19,13 +19,36 @@
...
+import re
+import six
+import sys
 import unittest
 
 
-class CoreTest(unittest.TestCase):
+class BaseTest(unittest.TestCase):
+  @contextlib.contextmanager
+  def assertRaisesFireExit(self, code, regexp=None):
```
jtratner: could move this base class somewhere else - wasn't sure how you wanted to organize it.

fire/fire_test.py
```diff
@@ -19,17 +19,17 @@
...
 
 
-class FireTest(unittest.TestCase):
-
+class FireTest(BaseTest):
   def testFire(self):
-    fire.Fire(tc.Empty)
-    fire.Fire(tc.OldStyleEmpty)
-    fire.Fire(tc.WithInit)
+    fire.Fire(tc.Empty, '')
```
jtratner: otherwise these error about `--ignore-files` from the nosetests command

fire/core.py
dbieber: I think non-zero, indicating that the trace ends in an error.

fire/core.py
```diff
@@ -141,7 +146,8 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    return None
+    # ape argparse by exiting out quickly
+    raise FireExit(0, None)
```
dbieber: Exiting quickly w/ code=0 sgtm. Let's still include the component_trace in the FireExit.

fire/core.py
```diff
@@ -127,7 +132,7 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    raise FireExit(1, component_trace)
+    raise FireExit(2, trace)
```
dbieber: trace is a module. You want component_trace.

fire/core.py
```diff
@@ -166,13 +172,16 @@ class FireExit(SystemExit):
 
   Contains the trace of the Fire program.
 
-  If not caught, then a FireExit will cause the program to exit with a non-zero
-  status code.
+  Attributes:
```
dbieber: Let's move the "Attributes" section into the docstring of __init__ and call it "Args:".

fire/core_test.py
dbieber: maybe let's make testutils.py. what do you think?

fire/core_test.py
```diff
@@ -66,12 +89,22 @@ def testInteractiveModeVariablesWithName(self, mock_embed):
     self.assertIsInstance(variables['trace'], trace.FireTrace)
 
   def testImproperUseOfHelp(self):
-    # This should produce a warning and return None.
-    with self.assertRaises(core.FireExit):
-      core.Fire(tc.TypedProperties, 'alpha --help')
+    # This should produce a help message
+    with self.assertRaisesFireExit(2, 'The proper way to show help.*Usage:'):
+      self.assertIsNone(core.Fire(tc.TypedProperties, 'alpha --help'))
```
dbieber: Drop the "assertIsNone"

fire/fire_test.py
```diff
@@ -19,17 +19,17 @@
 import fire
 from fire import test_components as tc
 from fire import trace
+from fire.core_test import BaseTest
```
dbieber: nit: we only import whole modules, never members of modules directly

```
from fire import testutils
class FireTest(testutils.BaseTest):
```

fire/fire_test.py
dbieber: It sounds like it was happening because sys.argv was being used. Previously it would just return None in the case of an error, but now it's SystemExiting.

Let's (at least for one test) mock argv rather than supplying a command string, so that the common case of no-command is represented in the tests.

fire/fire_import_test.py
```diff
@@ -12,16 +12,19 @@
...
+import fire
+
 
 class FireImportTest(unittest.TestCase):
   """Tests importing Fire."""
 
   def testFire(self):
-    fire.Fire()
+    with mock.patch.object(sys, 'argv', ['commandname']):
+        fire.Fire()
```
dbieber: nit: 2 spaces

fire/core_test.py
jtratner: sounds good - I'll do that

fire/core.py
jtratner: good call - fixed in f7c6a46 - I had started working on this branch and then rebased it over your code - my bad on missing it.  I updated the test to check the type of trace too

fire/fire_import_test.py
jtratner: whoops - translating between different style guides - do you have a set of lint settings that you want to apply to the repo? (or is there some google standard that can be added to travis runs)

fire/core.py
jtratner: sure - should I add class level variable like `trace = None` just to clarify to users that they can pull out the `trace` property to programmatically inspect on error?

fire/fire_import_test.py
dbieber: We have something like a pylintrc that I want to get checked into this project. The problem is it has a number of things that are internal specific, so it'll take some cleanup before I'm able to add it here. Definitely would like to get it into travis soon though.

I think it's pretty much this https://google.github.io/styleguide/pyguide.html, but I know there are some differences (e.g. FunctionNames are permitted.)

fire/core.py
jtratner: done in b4ab88f

jtratner:
I covered all your comments and fixed lint (I have to go figure out how to have a custom vim buffer settings for this repo so I stop forgetting to type `:set sw=2`)

fire/fire_import_test.py
jtratner: Might want to add link to that doc to CONTRIBUTING.md

jtratner:
New changes:
* separate testutils package
* updated test to check for type of FireTrace
* always has component trace on raise
* doc stuff

fire/core.py
```diff
@@ -132,7 +131,7 @@ def Fire(component=None, command=None, name=None):
     result = component_trace.GetResult()
     print(
         helputils.HelpString(result, component_trace, component_trace.verbose))
-    raise FireExit(2, trace)
+    raise FireExit(2, component_trace)
   elif component_trace.show_trace and component_trace.show_help:
     print('Fire trace:\n{trace}\n'.format(trace=component_trace))
     result = component_trace.GetResult()
```
dbieber: Think we should `raise FireExit(0, component_trace)` at lines 140 and 143 as well (the successful trace and successful help+trace cases)?

jtratner:
> 80 char lines

Ah - I've switched to the land of 99 characters ;)

fire/core.py
```diff
@@ -166,13 +171,18 @@ class FireExit(SystemExit):
...
-  status code.
+  Note:
+    Since this exception inherits from SystemExit/BaseException, you'll have to
+    explicitly catch it with `except SystemExit` or `except FireExit`.
   """
-
-  def __init__(self, status, data):
-    super(FireExit, self).__init__(status)
-    self.trace = data
+  def __init__(self, code, trace):
```
jtratner: Question, should we do:

```
__init__(self, code, trace)
```

which shadows trace in global namespace or

```
__init__(self, code, data)
```

like you had initially, so trace is not shadowed?

(found a pylintrc from apitools to run on this PR)

jtratner:
Do you want me to squash this down into a single commit?

fire/fire_test.py
```diff
@@ -288,12 +288,12 @@ def fn3(arg, **kwargs):
...
 
   def testTraceFlag(self):
-    self.assertIsInstance(
-        fire.Fire(tc.BoolConverter, 'as-bool True -- --trace'), trace.FireTrace)
-    self.assertIsInstance(
-        fire.Fire(tc.BoolConverter, 'as-bool True -- -t'), trace.FireTrace)
-    self.assertIsInstance(
-        fire.Fire(tc.BoolConverter, '-- --trace'), trace.FireTrace)
+    with self.assertRaisesFireExit(0, 'Fire trace:\n'):
+      fire.Fire(tc.BoolConverter, 'as-bool True -- --trace')
```
jtratner: only issue with switching to raising FireExit(0) on trace arg is it makes it a bit harder to work with interactively (e.g., can accidentally cause your shell to exit if you don't catch the exception) - but I guess you'd use the `--interactive` command line arg to do that instead?

dbieber: I know IPython will catch SystemExit exceptions and won't quit. Vanilla `python` quits though. This will also be an issue with `fire.Fire` even without the --trace flag.

dbieber:
> Do you want me to squash this down into a single commit?
No need; I'll squash+merge onto google/python-fire/z2017-03-21-fire-exit.

fire/core.py
dbieber: Let's avoid shadowing things from the global namespace.
How about we do `self.trace = component_trace` so that we're not introducing yet another name for traces.

Also can you link to the pylintrc you found?

jtratner: https://github.com/google/apitools/blob/master/default.pylintrc - wasn't actually that useful as a pylintrc, but it was good enough so I could set indent to '  ' and a line length and fix the lint errors for you :P

jtratner:
I think this is good for you to squash in now - I fixed the argument name.

jtratner:
should this just be repointed to merge into master and you can squash/merge from there?

dbieber:
Going to clean up some comments / lint things, so will merge onto z2017-03-21-fire-exit, then add a commit or two, then onto master from there.

PR #30, opened 2017-03-22
Fix failing tests on Windows due to hardcoded file paths

keatinge:
Some of the tests are currently failing on windows because the file paths are hardcoded for unix systems. For example, one test is checking for the presence of the below string:

`fire/test_components.py`

The problem is on the windows the actual string that appears is:

`fire\test_components.py`

This is a solved problem, [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join) is smart enough to figure out how to correctly join the files for whatever system the code is being executed on. The tests should now pass on both Windows and Unix systems.

### DIFF
--- a/fire/helputils_test.py
+++ b/fire/helputils_test.py
@@ -21,6 +21,7 @@ from fire import test_components as tc
 import six
 
 import unittest
+import os
 
 
 class HelpUtilsTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


fire/helputils_test.py
```diff
@@ -120,7 +121,7 @@ def testHelpClassNoInit(self):
       self.assertIn('Type:        type\n', helpstring)
     self.assertIn('String form: ', helpstring)
     self.assertIn('fire.test_components.OldStyleEmpty', helpstring)
-    self.assertIn('fire/test_components.py\n', helpstring)
+    self.assertIn(os.path.join('fire', 'test_components.py\n'), helpstring)
```
dbieber: Let's also drop the \n character here.

PR #31, opened 2017-03-23
Fix NameError: name 'x' is not defined in Argument Parsing example

Jackevansevo:
```python
fire.Fire(lambda obj: type(x).__name__) 
```
was causing a name error, changed `x` to `obj`


dbieber:
Good catch. Thanks for the fix.

Issue #29, opened 2017-03-22
Use fire without editing code directly


jtratner:
Currently you have to add something like this to the end of your file just to mess around with fire:

```
if __name__ == '__main__':
    fire.Fire()
```
and then run with

```
python path/to/code.py
```

But it'd be neat if instead you could just do

```
fire path/to/code.py
```

and that does equivalent of `exec` appending the ifmain statement to the end.

Not sure *how* to do it - I know kernprof does some magic to inject into builtins before exec'ing code.  (or maybe you get a code object back from exit) - but this would make fire even better as a debugging tool as well.

dbieber:
Cool idea.

Let's think through some of the behavior specifics.

- What happens if the file already has a main()? What if it's already using Fire or some other CLI library? Probably we don't want `__name__ == '__main__'` to be True when we run code.py, but inevitably if code.py has top-level code then it will be executed.
- Python supports 'python -m' for running a module. Do we want to support 'fire -m' as well?
- With regular Fire CLIs, we need to separate flags from the fire command with an isolated '--'. Here we could potentially allow the fire flags to preceed the filename. e.g. `fire -i code.py` or `fire -h code.py command`. Is this something we'd want?
- I have a design doc coming for updating tab completion; we'll have to think about how this would play alongside tab completion.

jtratner:
Great point on executing ifmain by accident.  This set me on to a neat idea - what if you could just pass in any fully qualified python module/function/class?

I snagged the AppEngine mapreduce's `for_name` to do the resolution and voila:

```
›› python fire/cli.py os.path basename a/b/c
c
›› python fire/cli.py os.path isfile a/b/c                                                                                                              
False
›› python fire/cli.py os.path exists a/b/c                                                                                                              
False
›› python fire/cli.py os.path exists fire/cli.py                                                                                                        
True
```

branch here with initial doodle - https://github.com/jtratner/python-fire/tree/run-on-any-fully-qualified-name

If we could get it to work so that `--interactive` ended up in the namespace of the module (rather than caller namespace) that'd make this incredibly useful.

I'll think about your other questions and get back to you later (have to get a deliverable together).

jtratner:
```
jtratner on c11553jtratner in python-fire/on run-on-any-fully-qualified-name
›› python fire/cli.py requests.get https://www.google.com - headers items
["Date", "Wed, 22 Mar 2017 21:29:50 GMT"]
["Expires", "-1"]
["Cache-Control", "private, max-age=0"]
[...]

PR #35, opened 2017-03-26
fire.py: Use fire on arbitrary file or module.

nealmcb:
Based on the discussion at [Use fire without editing code directly · Issue #29 · google/python-fire](https://github.com/google/python-fire/issues/29)

Of course, still needs better error handling, documentation, installation support, etc.


nealmcb:
It seems to pass the Travis CI on all the python 3 versions tested, but not python 2.  But in my own testing it works fine on python 2, and I haven't figured out where it's going wrong in `fire.fire_import_test.FireImportTest` with

    AttributeError: 'module' object has no attribute 'Fire'

Hints welcome :)

dbieber:
Thanks for starting this.

**Travis**
The immediate problem is that you've added a file named fire.py, so the test that tries to "import fire" ends up importing your file as the fire module, rather than the actual fire module. Using a different file name will get around this issue.

**Development workflow**
Like you say, there's other work to do before merging this to master (tests, documentation, etc). I'd be open to creating a feature branch for this if you prefer to work on this over multiple separate pull requests.

**Code review**
From a quick glance at the script, one issue I notice is that we shouldn't be modifying sys.argv. Instead, we'll want to modify the `Fire` function to accept a list for argv (currently it only accepts a string for the argument 'command', not a list).
I'll also try to get a pylintrc into the repo / checked by travis to make staying stylistically consistent easier.

nealmcb:
Thanks, @dberlin . That was my hunch for the import issue, but it makes me yet more curious about why it works in python 3.
A feature branch sounds helpful.

dbieber:
I changed the base to 'issue29-fire-without-edits'.

jtratner:
Another thing that's missing here is the ability to resolve to classes or functions within a module (tho I guess that's really just quibbling over addition or removal of an extra space)

dbieber:
> Classes or functions within a module (tho I guess that's really just quibbling over addition or removal of an extra space).

If they're within a module (e.g. `PIL.Image`) then it's just a matter of using a space instead of a dot.

If we want to support builtin classes/functions (e.g. `raw_input`, `type`, etc), then we need more than just find_module.

jack17529:
@dbieber is the PR done or something still remains?

dbieber:
I am OK w/ merging this to the issue29-fire-without-edits branch. Apologies for not merging sooner.

Some things that remain before merging to master:
- Integration into setup.py
- Documentation
- Removing requirement of -f/-m flag.
- lint compliance

nealmcb:
Thanks.

Also needed: tests!  Input on good testing approaches and frameworks for command-line functionality 
would be welcomed.

jack17529:
@nealmcb  I would love to work on tests as I have recently studied them, might require some help , can I take this up?

PR #38, opened 2017-03-27
Adds pylint script to .travis.yml

dbieber:


.travis.yml
```diff
@@ -6,6 +6,9 @@ python:
   - "3.6"
 before_install:
   - pip install --upgrade setuptools pip
+  - pip install --upgrade pylint pytest pytest-pylint pytest-runner
 install:
   - python setup.py develop
-script: nosetests --ignore-files=parser_fuzz_test.py --exclude=test_components_py3.py
+script:
+  - python -m pytest
```
jtratner: yay for pytest!

jtratner:
How would you feel about adding something like a Makefile so that people can run tests locally? I'm happy to put it up if you like.

I'm not sure how to replicate what you have on travis on my local machine tho (e.g., if I want to just run a test repeatedly or profile).

For example, running `python -m pytest` alone generates a ton of errors about other modules installed. (Whereas `python -m pytest fire` works quite well).

dbieber:
I'd be open to something like that, though I'm skeptical that adding a Makefile is necessary.
Would simply changing the .travis.yml file to say `python -m pytest fire` be sufficient?

PR #39, opened 2017-03-28
BUG: fix output printing on dict with no printable keys

jtratner:
Previously, outputs that generated dicts with all hidden keys would error with:

```
  File "/Users/jtratner/python-fire/fire/core_test.py", line 100, in testPrintEmptyDict
    core.Fire(tc.EmptyDictOutput, 'nothing_printable')
  File "/Users/jtratner/python-fire/fire/core.py", line 147, in Fire
    _PrintResult(component_trace, verbose=component_trace.verbose)
  File "/Users/jtratner/python-fire/fire/core.py", line 198, in _PrintResult
    print(_DictAsString(result, verbose))
  File "/Users/jtratner/python-fire/fire/core.py", line 221, in _DictAsString
    len(str(key)) for key in result.keys()
ValueError: max() arg is an empty sequence
```

This PR fixes that issue by filtering earlier in `_PrintResult` and adds a test case to cover the edge.

(I think I caught all the 2-spaces and lint errors this time haha)

### DIFF
--- a/fire/testutils.py
+++ b/fire/testutils.py
@@ -32,6 +32,15 @@ import six
 
 class BaseTestCase(unittest.TestCase):
   """Shared test case for Python Fire tests."""
+  @contextlib.contextmanager
+  def assertStdoutMatches(self, regexp):
+    """Asserts the context generates stdout matching regexp"""
+    stdout = six.StringIO()
+    with mock.patch.object(sys, 'stdout', stdout):
+      yield
+    value = stdout.getvalue()
+    if not re.search(regexp, value, re.DOTALL | re.MULTILINE):
+      raise AssertionError('Expected %r to match %r' % (value, regexp))
 
   @contextlib.contextmanager
   def assertRaisesFireExit(self, code, regexp=None):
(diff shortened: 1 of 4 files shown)


dbieber:
Thanks for catching and fixing this 👍 .

fire/testutils.py
```diff
@@ -32,6 +32,15 @@
 
 class BaseTestCase(unittest.TestCase):
   """Shared test case for Python Fire tests."""
+  @contextlib.contextmanager
```
dbieber: heads up I'm going to add a newline here

fire/testutils.py
```diff
@@ -32,6 +32,15 @@
...
+  def assertStdoutMatches(self, regexp):
+    """Asserts the context generates stdout matching regexp"""
```
dbieber: and a period here.

jtratner:
btw - this raised a question for me: ought fire to print trace (and help) messages to stderr rather than stdout?

dbieber:
Yeah, I think so. Same for the --help usage warning it prints too.

Issue #32, opened 2017-03-23
Improve Python Fire's tab completion


dbieber:
Please see [this early stage design document](https://docs.google.com/a/google.com/document/d/18rqDYmD7M9qtyrz08Ll-80jSrO_IaZhfx1hUx2aTllo/pub) for improving Python Fire's tab completion.

Comments on the design document are welcome! Let's have that discussion in this GItHub issue.
Contributors also welcome! :)

jtratner:
@dbieber - can you open up read access to everyone?

dbieber:
Try now

Ellmen:
I like this idea. I left a few thoughts in the doc. Not sure if this has been suggested before but it might make sense to have a "dev" version of the CLI with features like this and be able to ship a prod version with precomputed completions. I can take a stab at implementing this.

dbieber:
I'm definitely on board with being able to ship a version of a Fire CLI with precomputed completions 👍. I'm not sure I totally understand the distinction you're drawing between "dev" and "prod" though. In "prod" mode, generating new completions would be disabled? Is that the only distinction or are there others?

Ellmen:
Well another could be disabling the --completions flag in "prod". There could also be more verbose errors in "dev" although that would be outside the scope of the completions PR. Do you think it's too much overhead?

dbieber:
I want to make sure things keep working well in the simple case of:

```
if __name__ == '__main__':
  fire.Fire()
```

In particular it would be great if in this case you can get tab completions after having run your CLI only once (or, better, before you even run it at all! This is possible if the completion script runs your CLI with some flag when you hit tab, like your-command -- --get-completions. We'd have to make sure this doesn't execute code the user wasn't intending to execute though.).
Fire gets used during development a bunch, so it's important that completions can be invalidated once the system realizes they're no longer relevant.

Once a Fire CLI is shipped with completions, I don't think there's any problem with continuing to update the completions. If all the completions were included, these updates should be a no-op. I suppose I do see value in freezing the completions if there are completions the developer doesn't want to include.

One way to freeze completions (akin to your "dev" mode) would be having a config like fire.Fire(freeze_completions=True). Another approach would be to include something in the completions mapping that indicates that it's frozen (haven't thought this last bit through). I'm not sure how necessary completion freezing is, or if it's worth the complexity that it adds. Adding it to the list at #98 though as a possible config to support.

dbieber:
[...]

PR #40, opened 2017-03-28
Skip components we've already seen.

gaojiuli:
Inspired by built-in function `pkgutil.walk_packages()`

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -216,7 +216,7 @@ def _Commands(component, depth=3):
     Tuples, each tuple representing one possible command for this CLI.
     Only traverses the member DAG up to a depth of depth.
   """
-  
+
   if depth < 1:
     return
 
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -220,7 +220,7 @@ def _Commands(component, depth=3):
 
   for member_name, member in _Members(component):
     # Also skip components we've already seen.
-    if seen(p):
+    if seen(member_name):
         continue
     
     member_name = _FormatForCommand(member_name)
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -216,7 +216,6 @@ def _Commands(component, depth=3):
     Tuples, each tuple representing one possible command for this CLI.
     Only traverses the member DAG up to a depth of depth.
   """
-
   if depth < 1:
     return
 
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -210,11 +210,19 @@ def _Commands(component, depth=3):
     Tuples, each tuple representing one possible command for this CLI.
     Only traverses the member DAG up to a depth of depth.
   """
+  def seen(p, m={}):
+    if p in m:
+        return True
+    m[p] = True
+    
   if depth < 1:
     return
 
   for member_name, member in _Members(component):
-    # TODO: Also skip components we've already seen.
+    # Also skip components we've already seen.
+    if seen(p):
+        continue
+    
     member_name = _FormatForCommand(member_name)
 
     yield (member_name,)
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -222,7 +222,7 @@ def _Commands(component, depth=3):
 
   for member_name, member in _Members(component):
     # Also skip components we've already seen.
-    if seen(id(component)):
+    if _Seen(id(component)):
       continue
 
     member_name = _FormatForCommand(member_name)
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -196,10 +196,9 @@ def _FormatForCommand(token):
     return token.replace('_', '-')
 
 def _Commands(component):
-  seen = set()
-  return _CommandsHelper(component, seen)
-  
-def _CommandsHelper(component, seen):
+  return _CommandsHelper(component)
+
+def _CommandsHelper(component,seen=set()):
   seen.add(id(component))
 
   for member_name, member in _Members(component):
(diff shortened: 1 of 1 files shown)


dbieber:
Thanks for contributing!

I don't think this will produce the desired behavior though, since each call to _Commands will define it's own `seen` function with it's own `m`. Also the member_name is not sufficient to determine whether a component has been seen, since there can be different components with the same member_name.

For some context, one reason we would want to skip already seen components is to prevent an infinite loop when generating the completions (currently this is done via the `depth` parameter).

I went to prepare a minimal example:
```
import fire
x = {}
x['y'] = x
fire.Fire()
```
(but fire breaks on this because it tries to serialize x with json, which can't handle the circular reference :/)

So, here's a different small example:
```
import fire

class Sample(object):
  def __init__(self):
    self.prop = None

sample = Sample()
sample.prop = sample

fire.Fire()
```

`python example.py sample prop prop prop prop prop` is a perfectly valid command. But we don't want to run forever generating our completion script.
Maybe there's a change that we can make as part of #32 to allow for proper tab completion of this without an infinite length completion script :).

gaojiuli:
I have done a  an experiment.

```python
def test(a, m={}):
    print(id(test))

    def seen(p):
        if p in m:
            return True
        m[p] = True
        print(id(seen))
        print(id(m))
        print(m)

    seen(a)

test(1)
test(2)
test(3)
test(3)
```

I get the result:

```
140318639763240
140318617455336
140318639562120
{1: True}
140318639763240
140318617455608
140318639562120
{1: True, 2: True}
140318639763240
140318617454792
140318639562120
{1: True, 2: True, 3: True}
140318639763240
```

The result shows that each call test has the same seen() and the same m. 
Different components with the same member_name. So we could use id(component) instead.

dbieber:
> We could use id(component) instead.

Good thought.

Another thing we'll have to address is that we don't want "seen" to be a lifetime property of an object. Instead, we want to check whether we've seen an object before during the current top-level call to `_Commands`.

One way we might do this is with a `set` of ids and a helper function. Something like this:

```
def _Commands(component):
  seen = set()
  return _CommandsHelper(component, seen)
  
def _CommandsHelper(component, seen):
  seen.add(id(component))

  for member_name, member in _Members(component):
    if id(member) in seen:
      continue
    member_name = _FormatForCommand(member_name)
    yield (member_name,)
    
    if inspect.isroutine(member) or inspect.isclass(member):
      for completion in Completions(member):
        yield (member_name, completion)
      continue  # Don't descend into routines.
 
    for command in _CommandsHelper(member, seen):
      yield (member_name,) + command
  
  seen.remove(id(component))
```

gaojiuli:
I tried this:

```python
def _Commands(component):
    seen = set()
    return _CommandsHelper(component, seen=seen)


def _CommandsHelper(component, seen):
    seen.add(id(component))
    print(id(seen), seen)


_Commands("1")
_Commands("2")

_Commands("1")
_Commands("2")

```

result:
```
139921437490312 {139921437069920}
139921437490312 {139921437070032}
139921437490312 {139921437069920}
139921437490312 {139921437070032}

```
I tried this:

```python
def _Commands(component):
    return _CommandsHelper(component)


def _CommandsHelper(component, seen=set()):
    seen.add(id(component))
    print(id(seen), seen)


_Commands("1")
_Commands("2")

_Commands("1")
_Commands("2")
```
result:

```
139851851967624 {139851851547232}
139851851967624 {139851851547232, 139851851547344}
139851851967624 {139851851547232, 139851851547344}
139851851967624 {139851851547232, 139851851547344}
```

The second result is what i want.

dbieber:
I think the first of those two experiments is actually closer to what we want. Calling `_Command` multiple times should give the same result each time. It's _within_ a command that we want to avoid repeating component ids, since that's how we avoid the infinite loops I described.

I put together what I think is a complete implementation, though I haven't tested it:

```python
def _Commands(component, depth=3, avoid_cycles=True, used_ids=None):
  """Yields tuples representing commands for the component.
  
  Args:
    component: The component considered to be the root of the yielded commands.
    depth: The maximum depth with which to traverse the member DAG for commands.
        If depth is None, no depth constraints are imposed.
    used_ids: The set of the ids of the components already used in the command.
    avoid_cycles: If True, then commands will not repeat component ids they've
        already used.
  Yields:
    Tuples, each tuple representing one possible command for this component.
  """
  used_ids = used_ids or set()
  used_ids.add(id(component))

  if depth is not None and depth < 1:
    return
  next_depth = depth - 1 if depth is not None else None

  for member_name, member in _Members(component):
    if avoid_cycles and id(member) in used_ids:
      continue

    member_name = _FormatForCommand(member_name)
    yield (member_name,)
    
    if inspect.isroutine(member) or inspect.isclass(member):
      for completion in Completions(member):
        yield (member_name, completion)
      continue  # Don't descend into routines.
 
    for command in _Commands(member, used_ids, next_depth, avoid_cycles):
      yield (member_name,) + command
  
  used_ids.remove(id(component))
```

Calling `_Commands(component, depth=3, avoid_cycles=False)` should have the same behavior that the current `_Commands` implementation has, whereas calling `_Commands(component, depth=None, avoid_cycles=True)` should avoid repeating components within any command.

We'll have to add tests to verify this behavior before we're ready to check this or another implementation in.

What do you think?

gaojiuli:
Alright, I agree with you！I will keep an eye on this project。

PR #48, opened 2017-04-06
Issue 42 : Handled exception of ValueError and added exception message in the output.

saurabhkpatel:
**Example:**

```
import fire
a="fire"
x = {}
x['y'] = x
fire.Fire()
```

**output:**

```
a:    fire
fire: <module 'fire' from '/Users/srpatel/github/python-fire/fire/__init__.pyc'> [ Error:<module 'fire' from '/Users/srpatel/github/python-fire/fire/__init__.pyc'> is not JSON serializable]
x:    {'y': {...}} [ Error:Circular reference detected]
```


dbieber:
saurabh asked that I delete this pull request and he will open a new one as a different user.

Also let's not show the error ("[ Error:Circular reference detected]"), simply showing {'y': {...}} I think is sufficient.

PR #50, opened 2017-04-06
Added .gitignore file.

saurabhkpatel:
Used the standard one: https://github.com/github/gitignore/blob/master/Python.gitignore
Added support for intelliJ/pycharm .idea folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,100 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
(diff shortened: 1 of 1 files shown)


dbieber:
Let's check this in. Can you add a newline at the end first, please?

Some commentary:
- I want to avoid having every contributor adding their own ignore rules to google/python-fire's .gitignore file. See the section on [explicit repository excludes](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes) for how to additionally ignore files that only apply to you.
- google/python-fire's .gitignore file should only have rules that apply to a very large number of developers. Using GitHub's provided Python.gitignore file sounds good to me. I'm OK w/ adding .idea at the end since PyCharm is common (can you fix the capitalization btw?), though want to be careful not to set a precedent of modifying the .gitignore for individual developers.

PR #54, opened 2017-04-08
Write help and trace messages to stderr

jtratner:
Implements what we discussed on a separate PR, that trace and help output should go to stderr.

Updates `assertStdoutMatches` to capture stdout and stderr in one go (I figured it'd be easier to do that then have to use two with statements to check).  I used `stderr` and `stdout` for arg names to make it easier to read in tests.


jtratner:
(fixing the lint errors now)

dbieber:
Here's the stuff from our internal linter that Travis doesn't catch. :/. If you know how to make Travis catch these, lmk...

```
ERROR: Lint errors found:
Missing Yields section in docstring, or wrong choice of Returns/Yields [g-doc-return-or-yield]
        //depot/google3/third_party/py/fire/testutils.py:24

One-line docstring summary should end with ".", "?", or "!" [g-short-docstring-punctuation]
        //depot/google3/third_party/py/fire/test_testutils.py:1

Need 2 blank lines before top level class [g-wrong-blank-lines]
        //depot/google3/third_party/py/fire/test_testutils.py:8

One-line docstring summary should end with ".", "?", or "!" [g-short-docstring-punctuation]
        //depot/google3/third_party/py/fire/test_testutils.py:10

Analyzed 7 file(s). 4 new error(s) and 0 non-actionable issue(s) found.
```

jtratner:
```
Missing Yields section in docstring, or wrong choice of Returns/Yields [g-doc-return-or-yield]
        //depot/google3/third_party/py/fire/testutils.py:24
```

how do you usually document context managers?

jtratner:
Nvm - saw that you added a `Yields` section below so I added something equivalent - should all pass now I think

fire/testutils.py
```diff
@@ -58,21 +83,15 @@ def assertRaisesFireExit(self, code, regexp=None):
     """
     if regexp is None:
```
dbieber: as long as you're in here, this can be the default value to the fn.
`def assertRaisesFireExit(self, code, regexp='.*'):`

fire/testutils.py
```diff
@@ -34,14 +34,41 @@ class BaseTestCase(unittest.TestCase):
...
-    if not re.search(regexp, value, re.DOTALL | re.MULTILINE):
-      raise AssertionError('Expected %r to match %r' % (value, regexp))
+  def assertOutputMatches(self, stdout='.*', stderr='.*', capture=True):
+    """Asserts that the context generates stdout and stderr matching regexps.
+
+    Args:
+      stdout (str): regexp to match against stdout (None will check no stdout)
+      stderr (str): regexp to match against stderr (None will check no stderr)
+      capture (bool, default True): do not bubble up stdout or stderr
+    Note:
```
dbieber: Let's move the note above the args section.

```
"""Asserts that the context generates stdout and stderr matching regexps.

Note: If wrapped code raises an exception, stdout and stderr will not be
checked.

Args:
...
```

fire/test_testutils.py
```diff
@@ -0,0 +1,53 @@
...
+      with self.assertOutputMatches(stderr=None):
+        print('blah', file=sys.stderr)
+
+    with six.assertRaisesRegex(self, AssertionError, 'stderr:'):
+      with self.assertOutputMatches(stdout='apple', stderr=None):
+        print('apple')
+        print('blah', file=sys.stderr)
+
+  def test_correct_ordering_of_assert_raises(self):
+    # check to make sure FireExit tests are correct
```
dbieber: nit: `Check to make sure FireExit tests are correct.`

fire/test_testutils.py
```diff
@@ -0,0 +1,53 @@
...
+    with self.assertOutputMatches(stdout='confirmed.*not working'):
+      with self.assertRaises(ValueError):
+        print('''Yep, we've confirmed it.\nThe receiver is not working. :(''')
```
dbieber: Use double quotes since you're quoting a single quote and it fits on one line.
Also, what is the meaning of this string?

fire/test_testutils.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Test the test utilities for Fire's tests."""
+
+from __future__ import print_function
```
dbieber: Also include
```
```

fire/test_testutils.py
```diff
@@ -0,0 +1,53 @@
...
+
+import six
+
+from fire import testutils
+
+
+class TestTestUtils(testutils.BaseTestCase):
+  """Let's get meta."""
+
+  def test_no_check_on_exception(self):
```
dbieber: We've been using testNames instead of test_names so far. Let's keep that for consistency.

fire/testutils.py
```diff
@@ -34,14 +34,41 @@ class BaseTestCase(unittest.TestCase):
...
-    with mock.patch.object(sys, 'stdout', stdout):
-      yield
-    value = stdout.getvalue()
-    if not re.search(regexp, value, re.DOTALL | re.MULTILINE):
-      raise AssertionError('Expected %r to match %r' % (value, regexp))
+  def assertOutputMatches(self, stdout='.*', stderr='.*', capture=True):
+    """Asserts that the context generates stdout and stderr matching regexps.
+
+    Args:
+      stdout (str): regexp to match against stdout (None will check no stdout)
```
dbieber: Is this way of documenting types common?
So far we've been spotty about when we include them, and we always put them after the colon when we do. `stdout: (str) regexp...`

jtratner: I was following this guide :) - http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

fire/test_testutils.py
jtratner: I just wanted to make sure it made something multiline - text doesn't matter at all.

jtratner: I made it something clearer

jtratner:
covered all the nits except type change in docstring (to me looks like types should be before the colon)

fire/test_testutils.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (C) 2017 Google Inc.
```
dbieber: oh this should be testutils_test.py

dbieber:
> covered all the nits except type change in docstring (to me looks like types should be before the colon)

You may be right (I'll poke around some more code to see if others do that). Even if we do switch to that though, we'll want to do it all at once.
Interestingly the guide that http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html links to doesn't actually mention types in docstrings at all.

And can you rename test_testutils.py to testutils_test.py?

jtratner:
Sure - happy to make it consistent and do the rename 😊. Just my muscle
memory conflicting😆. Prob can get to it tomorrow.

On Fri, Apr 7, 2017 at 7:40 PM David Bieber <notifications@github.com>
wrote:

covered all the nits except type change in docstring (to me looks like
types should be before the colon)

You may be right (I'll poke around some more code to see if others do
that). Even if we do switch to that though, we'll want to do it all at once.
Interestingly the guide that
http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
links to doesn't actually mention types in docstrings at all.

And can you rename test_testutils.py to testutils_test.py?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/google/python-fire/pull/54#issuecomment-292689435>, or mute
the thread

jtratner:
made the remaining changes @dbieber

jtratner:
I'm mystified by the build failure - for some reason pip on travis grabs IPython 6.0 (which is incompatible with 2.7), even tho locally same version of pip correctly grabs IPython 5.3.

Do you want to add an if statement for 2.7 that specifies IPython <6.0?  Not sure how to fix

dbieber:
Short term let's check in `ipython<6.0` as in #58 so we can get this in.

PR #55, opened 2017-04-09
WIP : Fix Part 1 of issue 51, uses os.path.basename(sys.argv[0])

saurabhkpatel:
Need help for creating the test case for this change.

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -54,6 +54,7 @@ from __future__ import print_function
 
 import inspect
 import json
+import os
 import pipes
 import shlex
 import sys
@@ -100,7 +101,7 @@ def Fire(component=None, command=None, name=None):
   # Get args as a list.
   if command is None:
     # Use the command line args by default if no command is specified.
-    name = name or sys.argv[0]
+    name = name or os.path.basename(sys.argv[0])
     args = sys.argv[1:]
   else:
     # Otherwise use the specified command.


dbieber:
To test a specific argv you can use mock.patch as shown [here](https://github.com/google/python-fire/blob/master/fire/fire_test.py#L35).
`    with mock.patch.object(sys, 'argv', ['progname']):`

I would add the test to core_test.py, maybe as a second test after `testFire` named `testFireDefaultName`.
The test could use a filepath for sys.argv[0], include '--', '--trace' as args, and then do something like check that the basename shows up in the output.

saurabhkpatel:
@dbieber  thanks for your suggestions :)
Never used mock in the python but this is the opportunity to learn from the community :)

I would write a new test method in core_test.py. My thoughts are similar like this.

```
  def testFireDefaultName(self):
    with mock.patch.object(sys, 'argv',
                           ['/github/python-fire/fire/base_filename.py']):
      output = fire.Fire(tc.Empty)
     assertIn("base_filename.py", output)    # would it work? 
```
I am not sure how to gather print output as return of `fire.Fire(tc.Empty)`.

Any suggestions?

dbieber:
Once #54 is checked in you can use `with self.assertOutputMatches` to check the output of something like `fire.Fire(tc.Empty)`. That should be checked in ~tomorrow.

dbieber:
^^ #54 is checked in now

dbieber:
Closing. Superseded by #59.

Issue #57, opened 2017-04-11
travis ci failing with advent of ipython 6.0.0rc1


nealmcb:
ipython 6.0.0rc1 was released 7 hours ago:

* https://github.com/ipython/ipython/releases

Now when building fire, travis CI finds that version, which doesn't support Python 2.7, and complains, e.g. at https://travis-ci.org/nealmcb/python-fire/jobs/220816890

```
...
Processing dependencies for fire==0.1.0
Searching for ipython
Reading https://pypi.python.org/simple/ipython/
Downloading https://pypi.python.org/packages/76/46/c505c48d264715136edde1a1960f9158133e34c03bbfe4c955dbc5948ce4/ipython-6.0.0rc1.tar.gz#md5=e57acb5fabd9f18b739de35a1f00eac5
Best match: ipython 6.0.0rc1
Processing ipython-6.0.0rc1.tar.gz
Writing /tmp/easy_install-Aw9UOD/ipython-6.0.0rc1/setup.cfg
Running ipython-6.0.0rc1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-Aw9UOD/ipython-6.0.0rc1/egg-dist-tmp-JVscrS
IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
...
error: Setup script exited with 1
The command "python setup.py develop" failed and exited with 1 during .
```

Time for a dependency change. For a lot of projects....

nealmcb:
Looks like `'ipython<6.0'` in setup.py will do the trick.

jtratner:
what's strange is that pip is supposed to ignore it (according to IPython's readme). I'm wondering if the IPython project messed up the trove classifiers or pip did something wrong

Carreau:
Sorry about that. I'm confused how you get rc1 without using the --pre flag. Or am I missing a configuration option. We'll improve the readme and the docs for downstream projects. Thanks a lot for catching this an warning us upstream.




> I'm wondering if the IPython project messed up the trove classifiers or pip did something wrong

The trove classifiers are purely informal. We (the ipython team) had to send patch upstream to Pip, Warehourse, and PyPI, plus modify some peps, in order for the `requires_python > 3.3` metadata to be understood by setuptools/pip.. etc. We likely missed something in setuptools.

jtratner:
Thanks @carreau - there are a bunch of moving parts here to take care of
[...]

Issue #7, opened 2017-03-05
Remove `IPython` dependency from `fire.Fire` or improve load times some other way


ubershmekel:
`import IPython` takes about 0.5 seconds on my machine to import and is the longest part of running simple `fire` scripts at the moment. Launching a python script and printing something takes 0.15 seconds so the total with `import fire` becomes about 0.65 seconds. 

It's only used in `_EmbedIPython` and `inspectutils.py` so maybe it can be optional for command line apps and the inspection part ported over (or placed in a different smaller package)? Would such a pull request be useful?

dbieber:
I agree that this is an important thing to improve, and it deserves some discussion first.

As you point out, IPython only used in two places:
- inspectutils.py only uses a single function from IPython (IPython.core.oinspect.Inspector().info), so maybe we could just strip that one function out of IPython or write an equivalent function from scratch. I would want permission from the IPython folks before doing the former.
- interact.py uses IPython for the --interactive flag, but this should really be optional. You'll notice the unused _EmbedCode function in interact.py. (https://github.com/google/python-fire/blob/master/fire/interact.py#L89) The reason that's there is to eventually enable Fire to be used without IPython as a required dependency.

The first course of action I'm leaning toward is to lazily import IPython in the two places it's used. If we do this then the lazy import (and the corresponding 0.5 second delay) will only trigger 1) when --interactive mode is used, and 2) when generating usage strings. We'll need to add the appropriate annotations to keep the linter happy since we'll be violating some PEP by lazy importing.

The second related thing I think we should do is make IPython an optional dependency for interact.py. If the user doesn't have IPython, then we should fall back to using `code` (a builtin module) instead. For this, we'll have to figure out a good way to make sure tests cover both of these situations. And I'm sure we'll be making the linter even more unhappy.

Thirdly, it would be good to remove inspectutils.py's dependency on IPython altogether. As noted above, we have two options for doing this (rewrite the info function or get permission from IPython to just take it).

Once all this is done, we could remove IPython from setup.py altogether or list it as an optional suggested dependency, if such a thing exists.

Thoughts?

fruch:
My vote is for optional usage of ipython as a subpackage, like `pip install fire[ipython]` only for the interactive mode to work. The default should be lighter. Most of my users won't ever use IPython.

As for the borrowing code from other IPython, that's also a good idea (maybe later it can become a share inspection of package)

tmr232:
Adding IPython devs to ask for their take on it .
[...]

PR #58, opened 2017-04-11
ipython<6.0 for #57; fix broken links in guide.md

nealmcb:
As @jtratner notes, I'm not sure about the best way to address the travis CI issues.
But I just ran into that while fixing the broken links in guide.md.

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -699,12 +699,12 @@ The complete set of flags available is shown below, in the reference section.
 
 | Using a CLI    | Command                    | Notes
 | :------------- | :------------------------- | :---------
-| [Help](doc/using-cli.md#help-flag) | `command -- --help` |Show help and usage information for the command.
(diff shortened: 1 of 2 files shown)


jtratner:
This seems like the right choice for "make it work for people to pip install on python 2.7" :) and can revisit later (or add an if statement based on python version)

dbieber:
On the docs links: thanks -- I also fixed those internally yesterday but hadn't pushed out the commit yet.
On the ipython<6 fix: that's unfortunate. I think what you did is good but hopefully we can find a better solution before doing a 0.1.1 release. Possible better solutions:
1: [eliminate ipython as a required dependency](https://github.com/google/python-fire/issues/7#issuecomment-284266940),
2: "fix" setuptools so it installs a 2.7 compatible version by default
3: maybe there's a way to specify the conditional dependency explicitly. not sure...

Will merge your PR in tomorrow.

PR #59, opened 2017-04-12
Fix Part 1 of issue 51, uses os.path.basename(sys.argv[0]) and Added …

saurabhkpatel:
…test case.

### DIFF
--- a/fire/fire_test.py
+++ b/fire/fire_test.py
@@ -43,6 +43,19 @@ class FireTest(testutils.BaseTestCase):
     self.assertEqual(fire.Fire(tc.OldStyleWithDefaults, 'double 2'), 4)
     self.assertEqual(fire.Fire(tc.OldStyleWithDefaults, 'triple 4'), 12)
 
+  def testFireDefaultBaseName(self):
+    with mock.patch.object(sys, 'argv',
+                           ['/github/python-fire/fire/base_filename.py']):
+      # positive case
+      with self.assertOutputMatches(stdout='Usage:       base_filename.py',
+                                    stderr=None):
+        fire.Fire(tc.Empty)
+      # negative case
+      with self.assertRaises(AssertionError):
+        with self.assertOutputMatches(stdout='Usage:fire/base_filename.py',
+                                      stderr=None):
+          fire.Fire(tc.Empty)
+
   def testFireNoArgs(self):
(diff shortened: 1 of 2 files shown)


fire/fire_test.py
```diff
@@ -43,6 +43,19 @@ def testFire(self):
     self.assertEqual(fire.Fire(tc.OldStyleWithDefaults, 'double 2'), 4)
     self.assertEqual(fire.Fire(tc.OldStyleWithDefaults, 'triple 4'), 12)
 
+  def testFireDefaultBaseName(self):
+    with mock.patch.object(sys, 'argv',
+                           ['/github/python-fire/fire/base_filename.py']):
```
dbieber: Let's use something like `os.path.join('python-fire', 'fire', 'base_filename.py')` so that we don't assume unix vs windows (/ vs \\). [Also dropping 'github' from the path.] It's not important if it's an absolute or relative path, either's fine.

fire/fire_test.py
```diff
@@ -43,6 +43,19 @@ def testFire(self):
...
+      # positive case
+      with self.assertOutputMatches(stdout='Usage:       base_filename.py',
+                                    stderr=None):
+        fire.Fire(tc.Empty)
+      # negative case
+      with self.assertRaises(AssertionError):
+        with self.assertOutputMatches(stdout='Usage:fire/base_filename.py',
```
dbieber: 1. In the negative case I don't think the string would be `'Usage:fire/base_filename.py'`, it would probably be
```
'Usage:       fire/base_filename.py'
```
We can either use \s+ to be invariant to the number of spaces. Or, in the negative case, we can drop 'Usage:' altogether to assert that 'fire/base_filename' doesn't appear at all. Or (my preferred option) we can drop the negative case altogether, since the positive case already asserts that the full path isn't showing up in the usage section.

2. I think it's an awkward construction to put assertOutputMatches inside of an assertRaises. (E.g. what if a different assertion fails (not that there is one) inside the assertOutputMatches context?) If we ever do want this, we'll want to make a helper 'assertNotOutputMatches` or something like that.

---------

I think we can safely drop the negative case altogether since we're already asserting there's space before 'base_filename' in the positive case.

fire/fire_test.py
saurabhkpatel: done

PR #65, opened 2017-04-15
Bug Fix : name was not setting if command is set

saurabhkpatel:


dbieber:
There's one failing test.
https://travis-ci.org/google/python-fire/jobs/222276022

```
=================================== FAILURES ===================================
_______________________ FireTest.testTabCompletionNoName _______________________
self = <fire.fire_test.FireTest testMethod=testTabCompletionNoName>
    def testTabCompletionNoName(self):
      with self.assertRaises(ValueError):
>       fire.Fire(tc.NoDefaults, '-- --completion')
E       AssertionError: ValueError not raised
fire/fire_test.py:328: AssertionError
```

Previously this would be a value error because no name was supplied. Now sys.argv[0] is being used as the name.

saurabhkpatel:
@dbieber do you recommend to remove this test case now? I am planning to remove this test-case.
As we know that by default we are assigning name to it.

dbieber:
Yep, I think removing the test sounds good. The other option would be modifying the test to check that completion works appropriately even when no name is specified.

saurabhkpatel:
Adjusted test case.

Issue #74, opened 2017-05-23
Support the option to provide a user-defined output formatter


monkut:
On trying out fire, one of the first things I wanted was a method to define an output formatter.

At the moment it appears there is a single static formatter defined in `_PrintResult()`.

Consider making an option in Fire:

`fire.Fire(OBJECT, formatter=myformatter)`

> at the moment this is a "formatter" and "outputter".  format and output should probably be separated...

dbieber:
Thanks for raising this issue.

Some brainstorming around the possible ways to approach this:
1. Pass in a formatter (as you suggested)
- Do we want a way for the formatter to fall back to the built in serializer?
2. Allow Fire's serialization to be silenced.
- Since the result is returned in the output of Fire already, the client can handle the serialization themselves.
3. Let objects implement something like __str__ or __firestr__ which fire's builtin serialization method will use.
4. Improve Fire's built in serializer, but don't allow custom serializers.
- There's definitely value in consistency. E.g. custom serialization probably wouldn't be compatible with piping outputs of one fire command as inputs to another (this feature may just be a pipe dream though :P).

Related note:
- One potential upcoming change to the serializer is that containers made purely of literals will be handled differently than containers with non-literals.
As an example, a dict of literals will be serialized in the current way (one key-value pair per line), whereas a dict with functions in it may get shown as a usage string, similar to how an object with properties is shown.

dbieber:
Objects that implement a custom `__str__` function will now be serialized with that! (New in v0.2.0.)

beasteers:
(sorry to comment on an old closed issue, I just didn't want to make issue duplicates)

@dbieber I do love that fire uses the `__str__` function of the object, but I still think there's utility in being able to pass a custom formatter as well.

Basically, If I want to format a nested dictionary so that the second level wasn't a single line dict and was, for example indented json or yaml (or something completely different, like filter certain keys, idk), I don't want to have to subclass `dict` and have that live in the core class/function just so I can provide a custom `__str__` formatter. 

If you're using basic functions with fire, you can just use a wrapper function to do what you need, but you can't easily do that with object methods especially once you start going down multiple levels of attributes (any recursive class wrapper would have to imitate attributes and `__class__`, `__dict__`, `__name__`, `__doc__` and all that that fire leverages and ughhh).

[...]

PR #75, opened 2017-05-23
Simple implementation of allowing externally defined formatters

monkut:
I noticed there is a TODO in `_PrintResult` to provide better handling of output formatting, but this is a quick an easy way to allow end-users to provide their own defined formatters.

For your consideration.

In response to raised issue #74

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -155,7 +155,7 @@ def Fire(component=None, command=None, name=None, formatter=None):
     raise FireExit(0, component_trace)
   else:
     result = component_trace.GetResult()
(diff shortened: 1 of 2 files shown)


dbieber:
FYI objects that defines a custom \_\_str\_\_ method are now serialized using that custom \_\_str\_\_ method.

dbieber:
Thanks for the PR. Closing as this is now obsolete.

We eventually added this functionality (years after your original PR) in https://github.com/google/python-fire/pull/345 
They key word arg is `serialize`.

PR #85, opened 2017-06-15
Add repr to output of argument parsing examples

mbarkhau:
This clarifies to me how things get parsed, more so than just having the type information.

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -567,26 +567,26 @@ of any argument we give it:
 
 ```python
 import fire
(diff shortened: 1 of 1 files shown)


mbarkhau:
I signed it!

mbarkhau:
Is there no desire for this PR? Should I close it?

dbieber:
One hesitation with this PR is that this example behaves differently in Python 2 vs 3.
Another is that it doesn't leverage Fire's auto-printing of results.
Otherwise, I appreciate the benefits of clarity that it provides.

Issue #81, opened 2017-06-02
Add licence to release


CJ-Wright:
Would it be possible to add the licence to the release?

dbieber:
Can you explain a bit more precisely what you'd like?
Currently we have a LICENSE.txt file in the repo and have the license set to "Apache Software License" in pypi.

CJ-Wright:
Yes those files are in the repo but not in your latest release. I was hoping for it in the release (conda-forge will track the licence in the package)

dbieber:
If convenient, can you point me to an example of a package that includes the LICENSE in the release, preferably one using setuptools, or submit a PR?

CJ-Wright:
I don't know what LICENSE you want to use so I can't really put in a PR.
Here is [scikit-beam](https://github.com/scikit-beam/scikit-beam/tree/v0.0.10) note the licence file.

Crispy33:
<a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/"><img alt="Creative Commons License" style="border-width:0" src="https://i.creativecommons.org/l/by-sa/4.0/88x31.png" /></a><br />This work is licensed under a <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License</a>.

CJ-Wright:
The issue is not the lack of license, it is that in your release on pypi doesn't have the `LICENSE` file. See [attrs](https://pypi.python.org/pypi/attrs) for a package which has the `LICENSE` file in the `tar.gz`

CJ-Wright:
Thank you! It seems to be fine, but I don't know if I'd know.

dbieber:
K, I just wanted give you a chance to object in case either the file name (LICENSE) or the location of the license in the tar (fire-0.1.1/LICENSE) weren't okay for conda.

CJ-Wright:
that works for me. Thank you again!

dbieber:
Done in 3e98ed5. We'll start having the LICENSE in the release with v0.1.2.

PR #88, opened 2017-07-22
Include LICENSE file in the output of "python setup.py sdist".

dbieber:
This is for issue #81.
Adds the LICENSE file to the release.

Note to future self: an alternative approach would be to use a MANIFEST.in file as in the https://github.com/python-attrs/attrs repo.


dbieber:
Closing in favor of https://github.com/google/python-fire/commit/3e98ed525eced05a5e78418f92770da28c71cf8e

PR #96, opened 2017-10-11
Fix for BinOp in args

### LINKED ISSUE #95: date string being parsed as integer math expression
Using fire 0.1.2...

$ python mycode.py my_function --daydate 2017-10-09 -> The daydate parameter of my_function is passed the string value '2017-10-09' as expected and desired.

$ python mycode.py my_function --daydate 2017-10-10 -> The daydate parameter of my function is passed the integer value 1997. Apparently '2017-10-10' is being parsed as an integer math expression? I'm not sure why anyone would ever want that, and I'm very sure that no one would expect that.

I understand that I can solve this by placing my argument value in both double and single quotes, but I'd really rather not have to do that.

mbarkhau:
Fixes #95
Don't evaluate top level ast.BinOp arguments, treat them as strings instead.

### DIFF
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -94,6 +94,9 @@ def _LiteralEval(value):
     SyntaxError: If the value string has a syntax error.
   """
   root = ast.parse(value, mode='eval')
+  if isinstance(root.body, ast.BinOp):
+    return value
+  
   for node in ast.walk(root):
     for field, child in ast.iter_fields(node):
       if isinstance(child, list):
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -95,7 +95,7 @@ def _LiteralEval(value):
   """
   root = ast.parse(value, mode='eval')
   if isinstance(root.body, ast.BinOp):
-    return value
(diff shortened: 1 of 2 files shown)


dbieber:
Thanks for submitting this.
Can you confirm it doesn't break parsing negative numbers?

mbarkhau:
I added an assertion, yes looks like negative numbers are still parsed as before to type int or float. I guess the expression must be `ast.UnaryOp` or something.

fire/parser.py
```diff
@@ -94,6 +94,9 @@ def _LiteralEval(value):
     SyntaxError: If the value string has a syntax error.
   """
   root = ast.parse(value, mode='eval')
+  if isinstance(root.body, ast.BinOp):
+    return value
```
dbieber: Let's raise a ValueError here instead, so that the logic of "treat this thing that doesn't parse as a string" all exists in one place (line 74).

mbarkhau: Ahh right, the fallback logic is to treat it as a string anyway. Good point.

PR #99, opened 2017-10-13
Do not treat arguments that start with '--' as string.

### LINKED ISSUE #84: Better handling of argument name typos
Using this minimal example of a Python-fire CLI:

```python
import fire

def f(my_arg):
    print(my_arg)

fire.Fire(my_arg)
```

This is the result I get:

```bash
➜  ~ python example.py --my-arg=hello
hello
➜  ~ python example.py --myarg=hello
--myarg=hello
```

As you can see, a typo in the argument leads to unexpected behaviour, which can be really hard to debug.

I would suggest never treating an argument matching `--[a-zA-Z0-9_-]+` as positional, without some sort of escaping by the user (perhaps similar to the escaping you require to avoid parsing ints, etc.).

greenmoon55:
Fixes #84

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -565,7 +565,8 @@ def _MakeParseFn(fn):
 
   def _ParseFn(args):
     """Parses the list of `args` into (varargs, kwargs), remaining_args."""
-    kwargs, remaining_kwargs, remaining_args = _ParseKeywordArgs(args, all_args, fn_spec.varkw)
+    kwargs, remaining_kwargs, remaining_args = \
+        _ParseKeywordArgs(args, all_args, fn_spec.varkw)
 
     # Note: _ParseArgs modifies kwargs.
     parsed_args, kwargs, remaining_args, capacity = _ParseArgs(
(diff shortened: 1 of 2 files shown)


fire/core.py
```diff
@@ -730,15 +732,17 @@ def _ParseKeywordArgs(args, fn_args, fn_keywords):
         if got_argument and (keyword in fn_args or fn_keywords):
           kwargs[keyword] = value
           skip_argument = not contains_equals and not is_bool_syntax
-          arg_consumed = True
+        else:
+          unconsumed_named_args.append(argument)
+        arg_consumed = True
```
dbieber: We can remove `arg_consumed` altogether (here and at 698), and replace the if at 739 with an `else`.

fire/core.py
```diff
@@ -684,6 +685,7 @@ def _ParseKeywordArgs(args, fn_args, fn_keywords):
     remaining_args: A list of the unused arguments from the original args.
```
dbieber: - Please update the Returns: section of the docstring.
- Also how about "remaining_kwargs" instead of unconsumed_named_args for consistency with the existing terminology?

greenmoon55:
Thanks! I'll look into it.

dbieber:
Heads up that some of the tests are failing. You can see the details here (https://travis-ci.org/google/python-fire/jobs/287352133) or by clicking "Details" next to the continuous-integration line after "Some checks were not successfull".

It looks like the majority of the failures are using the components tc.Kwargs or tc.MixedDefaults.

A good first place to investigate might be why this assert is failing [[1]](https://travis-ci.org/google/python-fire/jobs/287352133#L672) [[2]](https://travis-ci.org/google/python-fire/jobs/287352133#L759):

https://github.com/google/python-fire/blob/b5053bad7c92bb2ee73d99d924b09a48d1498d67/fire/fire_test.py#L344-L346

Here's the definition of MixedDefaults:

https://github.com/google/python-fire/blob/a0791c9d40565a36bc540a168a6c19670dd90c78/fire/test_components.py#L74-L83

---

For what it's worth, these components are also used in many of the failing tests:

https://github.com/google/python-fire/blob/a0791c9d40565a36bc540a168a6c19670dd90c78/fire/test_components.py#L56-L62

https://github.com/google/python-fire/blob/a0791c9d40565a36bc540a168a6c19670dd90c78/fire/test_components.py#L175-L184

fire/core.py
```diff
@@ -681,64 +683,72 @@ def _ParseKeywordArgs(args, fn_args, fn_keywords):
...
-            value = 'True'
-          elif keyword.startswith('no'):
-            keyword = keyword[2:]
-            value = 'False'
-          else:
-            value = 'True'
+  remaining_kwargs = []
+  remaining_args = []
+
+  if not args:
```
greenmoon55: I add a return statement here because pylint complains about "Too many nested blocks (6/5)". Not sure if I should do it this way.

greenmoon55:
@dbieber Thanks for the tip! I forgot to save the value corresponding to the keyword.

fire/fire_test.py
```diff
@@ -368,10 +368,15 @@ def testBoolParsingLessExpectedCases(self):
         fire.Fire(tc.MixedDefaults,
                   command=['identity', 'True', '10']), (True, 10))
```
dbieber: Another thing that may be worthwhile to test is the behavior in the presence of separators / chained function calls.

For example, with these components:

```
def example_component(arg1, kwarg2='default2'):
  def inner_component(kwarg3='default3'):
    return kwarg3
  return inner_component
```

```
def example_component(arg1, *varargs):
  def inner_component(kwarg3='default3'):
    return kwarg3
  return inner_component
```

What is the behavior of these calls?

fire.Fire(example_component, command=['value1', '-', '--kwarg3', 'value3']) == 'value3'
fire.Fire(example_component, command=['value1', '--kwarg3', 'value3'])

dbieber:
I'm going to go ahead and squash + merge. If you're up for adding those tests and/or making the nit fix, go ahead and do so in a follow up PR.

PR #100, opened 2017-10-18
Support for a flag to inject environment variables

### LINKED ISSUE #94: Support for environment variables as default
As I user I'd like to be able to specify some of my arguments as environment variables.

Quite often I have a single bash script performing many calls to a script written using python-fire. Many of the functions defined in the python script contain shared arguments which I am just effectively passing from my environment variables to each of the calls. Having a way to pass those through would be extremely useful.

## Desired behaviour:

Given:
```python3
$ cat run.py
import fire
def function_1(arg1, arg2=1):
    print(arg1, arg2)
fire.Fire(function_1)
```

I'd like:
```bash
$ arg1=arg1 python3.6 run.py -- --use_environment
arg1 1
$ arg1=arg1env arg2=arg2env python3.6 run.py arg1 arg2 -- --use_environment
arg1 arg2
$ arg1=arg1env arg2=arg2env python3.6 run.py arg1 -- --use_environment
arg1 arg2env
$ arg1=arg1env arg2=arg2env python3.6 run.py -- --use_environment
arg1env arg2env
$ arg1=arg1env python3.6 run.py -- --use_environment
arg1env 1
```

This would be extremely useful as then I'd be able to have the following:

```bash
export HOST=mydesiredhost
./machine_manager ping -- --use_environment
./machine_manager shutdown -- --use_environment
# No need to pass --host=mydesiredhost on each command
```

mariocj89:
Add a new option that will take defaults from environment variables if available.

The variable is looked up as the parameter and full uppercase for convenience.

The option use_environment can be used both as a flag or a value can be passed in which allows to set a prefix for the env variables to look for. (Check the tests)

I'm not totally convinced on how the variable is passed around the code. Open to suggestions.

Happy to add docs once we finalise all decisions.

Fixes #94

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -321,6 +321,7 @@ def _Fire(component, args, context, name=None):
   show_completion = parsed_flag_args.completion
   show_help = parsed_flag_args.help
   show_trace = parsed_flag_args.trace
+  use_environment = parsed_flag_args.use_environment
 
   # component can be a module, class, routine, object, etc.
   if component is None:
@@ -363,7 +364,7 @@ def _Fire(component, args, context, name=None):
         filename, lineno = inspectutils.GetFileAndLine(component)
 
         component, consumed_args, remaining_args, capacity = _CallCallable(
-            component, remaining_args)
+            component, remaining_args, use_environment)
 
         # Update the trace.
         if isclass:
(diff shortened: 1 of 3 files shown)


dbieber:
Thanks for the PR!
I will follow up with comments (and a draft of the configs plans / naming conventions) later.
Thanks for your patience.

mariocj89:
No thanks needed!

TomFaulkner:
Is this PR going to be merged? I'm looking forward to using this feature.

mariocj89:
I am happy to do any requested change :)

@dbieber let me know if there is anything to change

mariocj89:
I guess this is not going to happen

TomFaulkner:
It is a sad day.

PR #111, opened 2018-01-11
unicode() --> six.u() for Python 3 compatibility

cclauss:
__unicode()__ was removed from Python 3 because all strs are unicode.

Executing these lines in Python 3 would raise a NameError.

* https://pythonhosted.org/six/#six.u

### DIFF
--- a/fire/parser_fuzz_test.py
+++ b/fire/parser_fuzz_test.py
@@ -67,8 +67,8 @@ class ParserFuzzTest(testutils.BaseTestCase):
       raise
 
     try:
-      uvalue = unicode(value)
-      uresult = unicode(result)
+      uvalue = six.u(value)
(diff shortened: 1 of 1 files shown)


dbieber:
lgtm!, thanks

djizm:
What is 120:60:130:LG simplink up skaleing hdmi---------4/4/4/3/4/4...0hms
speakers
---------- Forwarded message ----------

dbieber:
Turns out `six.u` does not behave the same as `unicode` in Python 2. six.u expects a string literal as input, whereas unicode can accept other types like bool and converts them to unicode.
So when `value` is a bool, six.u errors.

dbieber:
Yeah, that looks right. Thanks.

Issue #56, opened 2017-04-10
Completions for fish shell


JelteF:
It would be really great to have completions generated for the fish shell: https://fishshell.com/docs/current/commands.html#complete

dbieber:
Thanks @Ellmen for adding fish completion support. This will go out in the next release. No eta for this just yet.

darrenburns:
Hi @dbieber, I notice that the fish completion support didn't go out in the latest release. Was this intentional? I found this repo and noted the fish completion support in the README on `master`, only to realise after trying to use it that there's no code in the latest release supporting this feature.

dbieber:
We're overdue for a release at this point, aiming to finish the help text rewrite before releasing. Fish support should be included in the next release. Apologies for the delay.

Issue #64, opened 2017-04-14
completion option for zsh


brucechou1983:
I'm using zsh 5.0.8 on OSX 10.11.6.

When I tried to run the generated completion bash for the basic example in README, I got the following error.

`_complete-calculatorpy:4: unrecognized modifier `C'`

I finally resolved this issue by replacing the `:` in the bash script with `: `.

An extra option for completion command will be helpful for zsh user.

dbieber:
> I finally resolved this issue by replacing the : in the bash script with :.

I don't follow -- you replaced a colon with a different colon?
It would be great to have zsh support, but I haven't looked into supporting it yet.
(I would also like a single completion script to work for all Fire CLIs, instead of having to generate one per CLI, but that's a bigger task than adding zsh support.)

llabusch93:
Hey guys,

is ZSH auto completion support planned?

Thanks for a followup.

loynoir:
Ran into same issue.
```txt
unrecognized modifier `C'
```

[Solved](https://github.com/google/python-fire/pull/320)

SedMek:
Thank you @loynoir for attempting to fix the issue, but it seems to me that your [PR](https://github.com/google/python-fire/pull/320) was closed, thus the problem of completion remains the same.

loynoir:
@SedMek 
- Ah, I signed CLA, but later merged [iln-sk](https://github.com/iln-sk) changes for some extra types, who seems not signed CLA, and PR got blocked by CLA bot.
- Plus I not using fire now, so I closed that blocked PR.

bukowa:
> @SedMek
> 
> * Ah, I signed CLA, but later merged [iln-sk](https://github.com/iln-sk) changes for some extra types, who seems not signed CLA, and PR got blocked by CLA bot.
> * Plus I not using fire now, so I closed that blocked PR.
> 
> [2faa993](https://github.com/google/python-fire/commit/2faa993496bed4b7eb7298c3e01c1c377f3fd5f6)

Does not work for me :(

maximehk:
The issue that triggered the `unrecognized modifier` error in ZSH should be fixed at head now. You can try and force re-installing the `python-fire` package.

`pip install --upgrade --force-reinstall python-fire`

GabLeRoux:
I confirm the zsh completion works on zsh. To test it, this is what I did:

```bash
git clone git@github.com:google/python-fire.git
cd python-fire
python setup.py install
```

Here's a simple example

```python
[...]

PR #122, opened 2018-05-17
Fish completion support

Ellmen:
Re: https://github.com/google/python-fire/issues/56

This adds support for completions for the fish shell. It adds an optional argument to the --completions flag to specify a shell to create completions for.

Usage looks like:
`widget -- --completion` or `widget -- --completion bash`
for bash completions or
`widget -- --completion fish`
for fish completions

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Please see [The Python Fire Guide](docs/guide.md).
 | [Help](docs/using-cli.md#help-flag) | `command -- --help` |
 | [REPL](docs/using-cli.md#interactive-flag) | `command -- --interactive` | Enters interactive mode.
 | [Separator](docs/using-cli.md#separator-flag) | `command -- --separator=X` | This sets the separator to `X`. The default separator is `-`.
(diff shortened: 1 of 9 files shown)


dbieber:
Thanks for this contribution!
Props for your thoroughness with docs and tests :)

dbieber:
Looking forward, we'd eventually like to replace our use of argparse in parser.py with fire.

When we do so, "completion" will become an argument of a function with default value 'None'. At that point '--completion' will be treated as `True`, '--completion bash' will be `'bash'`, and '--completion fish' will be `'fish'`. So, like in your code, we'll treat `True` as `'bash'`.

Ellmen:
Happy to help! Sounds good - it should be relatively easy to add zsh support the same way

dbieber:
That would be great -- #64 requests it.
I also welcome you to read the doc in #32 ([here](https://docs.google.com/document/d/1PvEK0nd_k3MxRjTuyH7lYvK4y4b2iYwickn2B56WomU/edit)) and let me know if you have thoughts on this or want to work on it.

PR #130, opened 2018-07-02
Fixes bug setting stories in guide.md

nfultz:
The example wass slightly broken, and always sets stories=1. Also changed the phew to be after a whole floor instead of every single step.


dbieber:
Thanks for the fix!

PR #131, opened 2018-07-24
WIP Adding __main__ script

nfultz:
This adds a  `__main__.py` to the package, with the goal of enabling one liners like:

```
$python -m fire urllib.parse - quote "foo bar"
foo%20bar
```


nfultz:
Let's close this in favor of #110 .

PR #135, opened 2018-08-18
Fix several warnings in the diff example

BoboTiG:
Fixes ResourceWarning unclosed file:
```python
examples/diff/diff.py:78: ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/tmpnby8pemm' mode='U' encoding='UTF-8'>
    self.tolines = open(tofile, 'U').readlines()
```
And DeprecationWarning 'U' mode is deprecated:
```python
examples/diff/diff_test.py::DiffTest::testUnifiedDiff
  examples/diff/diff.py:77: DeprecationWarning: 'U' mode is deprecated
    with open(fromfile, 'U') as f:
```

### DIFF
--- a/examples/diff/diff.py
+++ b/examples/diff/diff.py
@@ -74,9 +74,9 @@ class DiffLibWrapper(object):
 
     self.fromdate = time.ctime(os.stat(fromfile).st_mtime)
     self.todate = time.ctime(os.stat(tofile).st_mtime)
-    with open(fromfile, 'U') as f:
+    with open(fromfile) as f:
       self.fromlines = f.readlines()
-    with open(tofile, 'U') as f:
+    with open(tofile) as f:
       self.tolines = f.readlines()
 
   def unified_diff(self, lines=3):


BoboTiG:
I signed it!

dbieber:
Thanks for the PR, I’ll take a look this week

dbieber:
Thanks for fixing those warnings, this looks good.
For context, it's written the way it is because this is a direct translation of https://docs.python.org/2/library/difflib.html (section 7.4.5) to Fire.
I wonder if there's a way to push the changes upstream to the Python 2 documentation so we don't diverge from their example.

The context manager change lgtm.
For the open mode change, is there a way to do this without checking the version number with sys? E.g. does `six` have this mode? Or should we just use mode 'r' in all cases? I ask because putting the version check directly in the example sets us up for needing future changes later.

BoboTiG:
Thanks for the review.

I will open a PR for the Python documentation, good idea.
About the open mode, I would say we can go without any open mode as `r` is the default one. I am not aware a such a constant in `six`.

dbieber:
sgtm, let's remove the open mode.

> I will open a PR for the Python documentation, good idea.
Great, thanks. cc me, or drop a link to the PR in this issue once it's open?

BoboTiG:
Sure :)

BoboTiG:
Sorry, I will have time this week-end :)

dbieber:
Okay Sounds good

BoboTiG:
I aligned the script with the Python 3 version. WDYT?
You may have received notifications, but this is the BPO: https://bugs.python.org/issue34500.

dbieber:
lgtm,
will squash and merge this coming week.

dbieber:
All set, thanks!

Issue #108, opened 2018-01-02
No way to add shortened options (e.g. '-R/--recursive')


alexpilk:
Would it be possible to create short names for default boolean arguments?

For example, if you have the following function:
```python
def hello(greeting, uppercase=False):
    greeting = greeting.upper() if uppercase else greeting
    print(greeting)
```

It would be nice to have the following usages:
`python hello.py Hey! -u`  
`python hello.py Hey! --uppercase`

sdiepend:
This would indeed be a nice feature.

Ralph-Wang:
It would be a great feature.
but I'm concerned about the conflict of the first letter

alexshadley:
To deal with the issue of multiple args with the same letter, we could print a warning if a user attempts to use an ambiguous shortcut, i.e. one that could refer to multiple args.

I can go ahead and take a crack at this one!

dbieber:
Thanks for your implementation! Will respond more fully later. :)
I see you already closed your own pull request -- why is that?

alexshadley:
No problem!  While my code worked properly in Python 3.6, the Travis build failed for 2.7, 3.4, and 3.5.  I was going to re-open the pull once I'd worked those issues out.  By the way, how do you typically test compatibility across python versions?

dbieber:
Locally I test different python versions with virtualenvs and the commands listed in the travis file.
I have one virtualenv for Python 2 and one for Python 3 (rather than one for each of the four supported Python versions.)  `python -m pytest` and `pylint fire --ignore=test_components_py3.py,parser_fuzz_test.py`

I don't run these multi-version tests locally often though, and instead I usually rely on travis and/or the google-internal testing and presubmit system for more thorough testing of multiple versions.

The output from travis for your PR is here:
https://travis-ci.org/google/python-fire/builds/429276689?utm_source=github_status&utm_medium=notification

It looks like the failures are just linter failures (two line-too-long errors). The reason 3.6 passed and the others fail is that we don't run the linter for 3.6, whereas we do for the other versions.

alexshadley:
Thanks for the detailed respone.  Had some time to fix up the linting errors, pull request is reopened.  Let me know what you think!

dbieber:
@alexshadley's initial implementation is checked in now.

Some next steps before this issue is done:

- raise a FireError if the arg doesn't match any keyword
- give exact matches priority (e.g. if there are args named "a" and "apple" then -a should match "a", rather than being flagged as ambiguous)
[...]

Issue #102, opened 2017-11-28
Unexpected parsing behaviour for --date=2018-12-09 vs. --date=2018-12-10


martinvirtel:
Anything that can be interpreted as an arithmetics expression will be magically "calculated" by fire. This is very practical. But it bit me because I wanted to parse dates from command line parameters using fire, see below. 

Suggestion: accept type hints as a way to prevent this. A parameter with type hint "str" shoud not be "calculated". 

Test code:

```import fire

def test(string: str=None) :
    print(string)

if __name__=="__main__" :
    fire.Fire(test)
```

Will give you: 

```
> python ./test-parsing.py --string=2010-10-10
1990    # not interpreted as a string, but as 2010 minus 10 minus 10

> python ./test-parsing.py --string=2010-10-09
2010-10-09

```

dbieber:
Definitely agree that this is an issue.
In my view, it's also an issue even when there is no type annotation.
Expressions like 2010-10-10 should not be evaluated by Fire; only literals and collections of literals should be parsed by default. (Issue https://github.com/google/python-fire/issues/97)

danishprakash:
I would like to fix this @dbieber if nobody else is and while we're at it, do you already have an approach in mind for tackling this?

dbieber:
Yes, see my comment here: https://github.com/google/python-fire/issues/97#issuecomment-349153815

dbieber:
I think this was fixed at least by 0.2.1, though maybe it was earlier.

Issue #97, opened 2017-10-12
Only plain literals should be parsed


mbarkhau:
@dbieber correctly pointed out in #95 that there is a more general problem with the parsing code. This is a follow up issue to that issue which only dealt with the most common case of a single argument.

Here are some cases demonstrating
```bash
$ python3 -c "import fire.parser as p; v = p.DefaultParseValue('[1+1]'); print(type(v), v)"
<class 'list'> [2]
$ python3 -c "import fire.parser as p; v = p.DefaultParseValue('{key: 1+1}'); print(type(v), v)"
<class 'dict'> {'key': 2} 
```

Some options are:
1. Accept the existing behaviour.
2. Always parse the whole argument as a string by doing `raise ValueError` in ``fire.parser._LiteralEval`
3. Parse only the sub expressions as strings

For 2. and 3. I think these would be the valid types from the `ast` module. 

 - `Str`
 - `Name` (converted to `Str`)
 - `Num`
 - `List`
 - `Tuple`
 - `Dict`
 - `Set`
 - `UnaryOp`

dbieber:
First, something that confuses me:
The [literal_eval documentation](https://docs.python.org/2/library/ast.html#ast.literal_eval) says it is not capable of evaluating arbitrarily complex expressions, **for example involving operators** or indexing.

This made me wonder, why does ast.literal_eval("1+1") evaluate to 2? A: It doesn't.
So why does DefaultParseValue('[1+1]') give [2]? A: It doesn't.
Turns out this is a Python 2 vs Python 3 distinction. ast.literal_evak("1+1") does give 2 in Python 3 :(.

So if we can get a port of Python 2's literal_eval into Python 3, I think that resolves our issues and we can remove the BinOp special case altogether. I haven't investigated whether this exists yet.

--------
Aside:
Something to keep in the back of your mind as you think about this: it would be _amazing_ if there were a standard way of parsing literals from the command line that was consistent across languages. This way _if_ someone were to implement, say, Java Fire, _and_ if someone were to implement piping args from the command line into Fire CLIs, then Fire could become a mechanism for interoperating across languages.

dbieber:
The Python 3 documentation for literal_eval also suggests it won't parse expressions with operators.
[...]

PR #141, opened 2018-09-16
Issue #108: Shortcuts for boolean arguments

alexshadley:
In response to [this issue](https://github.com/google/python-fire/issues/108).  This adds a feature where a shortcut can be used for boolean arguments of functions.  For example the function

`def foo(bar=False):`

can be invoked as `foo -b` to set bar to True.  While developing this, I came across a couple concerns:

- In detecting whether or not arguments matched the pattern of a bool shortcut, I decided that python's regex library was the best way to solve this problem.  Is adding `re` to imports ok?
- If a naming collision occurs (see final assert of `testBoolShortcutParsing`) the program raises a FireError describing the ambiguity.  Is this a proper manner to use this error in?
- Currently, if a shortcut doesn't match any function argument, it is parsed as-is.  Should something else be done in this scenario?

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -60,6 +60,7 @@ import pipes
 import shlex
 import sys
 import types
+import re
 
 from fire import completion
 from fire import decorators
@@ -750,7 +751,7 @@ def _ParseArgs(fn_args, fn_defaults, num_required_args, kwargs,
 def _ParseKeywordArgs(args, fn_spec):
   """Parses the supplied arguments for keyword arguments.
 
-  Given a list of arguments, finds occurences of --name value, and uses 'name'
+  Given a list of arguments, finds occurrences of --name value, and uses 'name'
   as the keyword and 'value' as the value. Constructs and returns a dictionary
   of these keyword arguments, and returns a list of the remaining arguments.
 
@@ -767,6 +768,9 @@ def _ParseKeywordArgs(args, fn_spec):
     kwargs: A dictionary mapping keywords to values.
     remaining_kwargs: A list of the unused kwargs from the original args.
     remaining_args: A list of the unused arguments from the original args.
+  Raises:
+    FireError: if a boolean shortcut arg is passed that could refer to multiple
+        args
   """
   kwargs = {}
   remaining_kwargs = []
@@ -793,7 +797,7 @@ def _ParseKeywordArgs(args, fn_spec):
       contains_equals = '=' in keyword
       is_bool_syntax = (
           not contains_equals and
-          (index + 1 == len(args) or args[index + 1].startswith('--')))
+          (index + 1 == len(args) or args[index + 1].startswith('--') or re.match('^\-[a-z]$', args[index + 1])))
       if contains_equals:
         keyword, value = keyword.split('=', 1)
         got_argument = True
@@ -827,6 +831,16 @@ def _ParseKeywordArgs(args, fn_spec):
           remaining_kwargs.append(argument)
           if skip_argument:
             remaining_kwargs.append(args[index + 1])
+
+    # catch boolean shortcut args
+    elif re.match('^\-[a-z]$',argument):
+      keychar = argument[1]
+      potential_args = [arg for arg in fn_args if arg[0] == keychar]
+      if len(potential_args) == 1:
+        arg_consumed = True
+        kwargs[potential_args[0]] = 'True'
+      elif len(potential_args) > 1:
+        raise FireError("The argument '{}' is ambiguous as it could refer to any of the following arguments: {}".format(argument, potential_args))
 
     if not arg_consumed:
       # The argument was not consumed, so it is still a remaining argument.
--- a/fire/core.py
+++ b/fire/core.py
@@ -802,7 +802,7 @@ def _ParseKeywordArgs(args, fn_spec):
         elif len(potential_args) > 1:
           raise FireError("The argument '{}' is ambiguous as it could "
                           "refer to any of the following arguments: {}".format(
-            argument, potential_args))
+                              argument, potential_args))
 
       else:
         keyword = argument[2:]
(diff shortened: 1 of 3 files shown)


dbieber:
tl;dr I think we should support single-character flags (like your PR does), but not restrict them to bool-only values.

----

I was previously toying with the idea of allowing users to specify a flag (bool or otherwise) that is a prefix of exactly one valid parameter, but ultimately decided against that idea. Here is my thinking there (stronger points in bold):

Pros: 
- **This enables single-character flags**
- And is additionally even more general, handling the case where multiple params start with the same letter.

Cons:
- It enables people to accidentally pass args without knowing precisely what arg they're passing
- **Backwards compatibility for Fire CLIs**
  - For example, if there's a Fire CLI with param send_message, a user might make a script using that CLI using the flag `--send`. If the CLI author later wants to add a new param send_attachment, then the user's script becomes ambiguous and breaks.
- Backwards compatibility for Fire itself


-------

Your proposal differs in two significant ways: single-character-only instead of arbitrary prefixes, and bool-only instead of any type.

Single-character only still gets the most important Pro, which is supporting single-character flags.
Single-character only still has the most important Con too, which is that a single-character flag may be invalidated later.
Single-character only doesn't have as much of the final Con (backwards compatibility for Fire) since it promises a strict subset of what arbitrary-prefix does to Fire's users.
Similarly bool-only promises a strict subset of what any-type does.
However, bool-only deviates from one of Fire's current properties which is that you can pass any value (literal/container of literals) for any argument.


So my thought is that we should support single-character flags, but not restrict them to bool-only.

fire/core.py
```diff
@@ -852,6 +852,21 @@ def _ParseKeywordArgs(args, fn_spec):
...
+  return _IsSingleCharFlag(argument) or _IsMultiCharFlag(argument)
+
+
+def _IsSingleCharFlag(argument):
+  """Determines if the argument is a single char flag (e.g. '-a')"""
+  return re.match('^-[a-z]$', argument)
+
+
+def _IsMultiCharFlag(argument):
+  """Determines if the argument is a multi char flag (e.g. '-alpha')"""
```
dbieber: I think you meant '--alpha'

alexshadley: Yup!  Just pushed a fix

dbieber:
Oh, looks like you're actively working on this (excellent!), I'll hold off on looking through until you're settled.
If you're put off by having to go to travis to check the lint errors (you get there by clicking the red x next to your commit) I can clean those up on my end instead, just lmk.

fire/core.py
```diff
@@ -785,15 +789,27 @@ def _ParseKeywordArgs(args, fn_spec):
...
       got_argument = False
 
-      keyword = argument[2:]
+      keyword = ''
+      if _IsSingleCharFlag(argument):
+        keychar = argument[1]
+        potential_args = [arg for arg in fn_args if arg[0] == keychar]
+        if len(potential_args) == 1:
+          keyword = potential_args[0]
+        elif len(potential_args) > 1:
```
dbieber: we also need to handle the len(potential_args) == 0 case.

alexshadley: Should we also throw an error in this case?  My original assumption here was that if the flag doesn't match, we would assume it was an value for another flag.  Now that I think about it though this could cause confusing behavior.

dbieber: Yeah I think we should throw an error in this case.
If the argument looks like a flag (_IsFlag), it shouldn't be acceptable as a value.
[And I think we'll also broaden what _IsFlag accepts too in a separate change.]

alexshadley: Ok that sounds good to me, I'll add a branch and a test case for it

alexshadley:
Alright, I've run out of time tonight to work on this tonight.  I'll have time tomorrow night to finish up!

dbieber:
I'm going to squash and merge what you have so far, and you can put subsequent improvements in a new PR.

dbieber:
Some next steps:
- raise a FireError if the arg doesn't match any keyword
- give exact matches priority (e.g. if there are args named "a" and "apple" then -a should match "a", rather than being flagged as ambiguous)
- support equal signs with single-character args (-a=10 should be OK)

dbieber:
Let me know if you're interested in taking on these and if so, which ones

dbieber:
> raise a FireError if the arg doesn't match any keyword
Ah, looking back I see that's actually not how things are currently handled. Unparsed args are returned as part of remaining_kwargs.

alexshadley:
Yeah, I can start working on supporting equal sings in single-char args.

As far as raising a FireError if the arg doesn't match, how should we proceed?  I agree that it would be a nice feature, but as I looked into it further this seems to get tricky.  As far as I understand (correct me if I'm wrong here), we first attempt to instantiate the owning class with the arglist, and upon this failing we move on to calling the function.  When I attempted to raise an error on no applicable shortcuts, the error was just thrown whenever the class was parsed as the class had no args at all.

dbieber:
Ah was just about to post here, 1 sec:

dbieber:
I went ahead and put together a commit to add support for equal signs and giving exact matches priority already, I'll put it on GitHub now and drop a link here.

dbieber:
As for raising the FireError, I was mistaken about that. The correct behavior (e.g. in order to support --help properly) is to return unused kwargs in the remaining_kwargs entry of _ParseKeywordArgs. I do this in my commit.

dbieber:
https://github.com/google/python-fire/commit/0b0f2421ea8edcc11a80aadb1f574d687e94664f

alexshadley:
Ah, looks good then!  Any other areas of the project that need work done?

dbieber:
https://github.com/google/python-fire/issues/102 / https://github.com/google/python-fire/issues/97 looks like a fun one if you're up for AST traversals.

PR #143, opened 2018-10-06
Deal with classmethod and staticmethod

### LINKED ISSUE #113: How to deal with classmethod or staticmethod?
```
import fire
class Calculator(object):
  """A simple calculator class."""
  base = 10

  def __init__(self, x):
    self.x = x

  def double(self, number):
    return self.x * number

  @classmethod
  def test(cls, number):
    return cls.base * number

if __name__ == '__main__':
  fire.Fire(Calculator)
```
=== a.py  =====================================
If I want to call func test, I need to do like this:  python a.py --x 3 test 5 
So how to call classmethod or staticmethod without initialize the class （python a.py test 5 ）?

rozuur:
Fixes #113

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -575,6 +575,19 @@ def _GetMember(component, args):
 
   raise FireError('Could not consume arg:', arg)
 
+def _TryClassMethod(component, args):
+  """Reads first argument and if it is a classmethod returns it as a new component
+  """
+  if not args:
+    return component, 'class', args
+  attrs = inspect.classify_class_attrs(component)
+  class_methods = {a.name for a in attrs if a.kind == 'class method' or a.kind == 'static method'}
+  arg_name = args[0].replace('-', '_')
+  if arg_name in class_methods:
+    return getattr(component, arg_name), 'routine', args[1:]
+  else:
+    return component, 'class', args
+
 
 def _CallAndUpdateTrace(component, args, component_trace, treatment='class',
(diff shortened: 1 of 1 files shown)


fire/core.py
```diff
@@ -575,6 +575,19 @@ def _GetMember(component, args):
 
   raise FireError('Could not consume arg:', arg)
 
+def _TryClassMethod(component, args):
+  """Reads first argument and if it is a classmethod returns it as a new component
+  """
+  if not args:
+    return component, 'class', args
+  attrs = inspect.classify_class_attrs(component)
+  class_methods = {a.name for a in attrs if a.kind == 'class method' or a.kind == 'static method'}
```
dbieber: linter is complaining about line 579 and 584 being too long

fire/core.py
```diff
@@ -598,6 +611,10 @@ def _CallAndUpdateTrace(component, args, component_trace, treatment='class',
   """
   if not target:
     target = component
+
+  if treatment == 'class':
+    component, treatment, args = _TryClassMethod(component, args)
```
dbieber: Currently you have accessing and calling the class method as a single action.
We should split this into two actions so that the user can access the method without calling it.
This is important so that the user can get help about a class method. Getting help requires accessing the method but not calling it.

dbieber:
Closing as we've since fixed #113

PR #152, opened 2018-12-10
encode to utf-8

duke79:
I needed utf-8 and I found no other way than to include the complete repo in my project.
Does it look like a good idea in general?


duke79:
I signed it.

Borda (approved):

Issue #150, opened 2018-11-01
Bug: Chaining Function Calls broken?


borislopezaraoz:
This code from docs (simplified):

```
import fire

class Airport(object):
    def __init__(self, code):
        self.code = code

if __name__ == '__main__':
    fire.Fire(Airport)

```

Produces this error when trying to chain upper function (command also from docs):

```
python example.py --code=ALB code upper
Fire trace:
1. Initial component
2. Instantiated class "Airport" (example.py:3)
3. Accessed property "code"
4. Accessed property "upper"
5. ('The function received no value for the required argument:', 'self')

Type:        builtin_function_or_method
String form: <built-in method upper of str object at 0x1063f2a40>
Docstring:   Return a copy of the string converted to uppercase.

Usage:       example.py --code=ALB code upper SELF
             example.py --code=ALB code upper --self SELF
```

fire==0.1.3

dbieber:
Hmm, a quick check on my system shows `python example.py --code=ALB code upper` is correctly producing `ALB` for both Python 2.7 (ipython==5.5.0) and 3.5.4rc1 (ipython==7.2.0) for fire==1.3.0.

What version of Python are you on? Is ipython installed, and if so at what version? OS?

dbieber:
Fixed by 668007ae41391f5964870b4597e41493a936a11e.

PR #155, opened 2018-12-14
Check for self parameter

kinow:
Related to #150 , I also tried the example from `README.md`, and it failed to call the `str.upper()`. I tried the old tags, and none worked. So I guess one of the libraries (`oinspect`? `inspect`?), could have changed, and is now returning `self` in the list of arguments for the operation.

Tried with Python 2.7.15 (Ubuntu LTS), and Python 3.7 (Anaconda 5.3.1).

When that happens, the validation fails as the number of arguments provided (none), is different than the number of parameters required (1 at least, for `self`). I tried the most naive fix that I could think first, and it worked?

So added a few tests to confirm, executed the other tests, and it appears to be working? But definitely needs some better review.

Cheers
Bruno

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -705,6 +705,8 @@ def _ParseArgs(fn_args, fn_defaults, num_required_args, kwargs,
   parsed_args = []
   for index, arg in enumerate(fn_args):
     value = kwargs.pop(arg, None)
+    if arg == 'self':
+      continue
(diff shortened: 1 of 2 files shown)


dbieber:
Ah, I see I need to understand #150 before I can understand this issue.

dbieber:
Fixed by 668007ae41391f5964870b4597e41493a936a11e

PR #160, opened 2018-12-31
Fixed Typos on pylintrc

heysachin:


dbieber:
I believe we need the dot in requirements.txt as it refers to the current directory. The actual requirements are listed in setup.py.

Issue #157, opened 2018-12-27
Accessing namedtuple by field name


robertvazan:
When a command returns `namedtuple`, Python Fire renders it as a `tuple` or a `list` and allows indexing into it by offset. It would be better to present `namedtuple` as if it is a `dict` or an object and allow access to its members via field names.

dbieber:
Yes, agreed we should support this.

PR #161, opened 2019-01-02
Make namedtuple fields accessible by their name

floscha:
Hi, this PR makes the fields of namedtuple objects accessible by their name instead of indices as requested in #157. Also, I extended the test suite and made sure that no existing tests broke.

The exact behaviour is explained in the following:

Assuming a `namedtuple_example.py` script:
```python
from collections import namedtuple
import fire

Point = namedtuple('Point', ['x', 'y'])
p = Point(11, y=22)

if __name__ == '__main__':
  fire.Fire(p)
```

Before the PR:
```
python namedtuple_example.py 0  # 11
python namedtuple_example.py 1  # 22
python namedtuple_example.py x  # Throws an exception.
```

After the PR:
```
python namedtuple_example.py x  # 11
python namedtuple_example.py y  # 22
python namedtuple_example.py 0  # Throws an exception since p has no '0' field.
```

### DIFF
--- a/fire/inspectutils.py
+++ b/fire/inspectutils.py
@@ -204,7 +204,7 @@ def _InfoBackup(component):
   return info
 
 
-def Isnamedtuple(component):
+def IsNamedTuple(component):
   """Return true if the component is a namedtuple.
 
   Unfortunately, Python offers no native way to check for a namedtuple type.
--- a/fire/inspectutils.py
+++ b/fire/inspectutils.py
@@ -202,3 +202,25 @@ def _InfoBackup(component):
     pass
 
   return info
+
+
+def isnamedtuple(component):
+  """Return true if the component is a namedtuple.
+
+  Unfortunately, Python offers no native way to check for a namedtuple type.
+  Instead, we need to use a simple hack which should suffice for our case.
+  namedtuples are internally implemented as tuples, therefore we need to:
+    1. Check if the component is an instance of tuple.
+    2. Check if the component has a _fields attribute which regular tuples do
+       not have.
+
+  Args:
+    component: The component to analyze.
+  Returns:
+    True if the component is a namedtuple or False otherwise.
+  """
+  if not isinstance(component, tuple):
+      return False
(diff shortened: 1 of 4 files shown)


dbieber:
Thanks for the PR!
I’m not at a computer right now, so quick question:
Does namedtuple[0] work? If so, we should still support your final example rather than throwing an exception.

floscha:
Thanks for the quick feedback!
Yes, `namedtuple[0]` works indeed. I've now added some changes (plus respective unit tests) so that Fire behaves the following way:
```
python namedtuple_example.py x  # 11
python namedtuple_example.py y  # 22
python namedtuple_example.py 0  # 11
python namedtuple_example.py 1  # 22
python namedtuple_example.py -1  # 22
python namedtuple_example.py -2  # 11
python namedtuple_example.py -3  # Fails while printing the appropriate trace.
python namedtuple_example.py 2  # Same as above.
```

fire/inspectutils.py
```diff
@@ -202,3 +202,25 @@ def _InfoBackup(component):
     pass
 
   return info
+
+
+def Isnamedtuple(component):
```
dbieber: Use casing "IsNamedTuple"

fire/inspectutils.py
```diff
@@ -202,3 +202,25 @@ def _InfoBackup(component):
...
+  Args:
+    component: The component to analyze.
+  Returns:
+    True if the component is a namedtuple or False otherwise.
+  """
+  if not isinstance(component, tuple):
+    return False
+
+  has_fields = bool(getattr(component, '_fields', None))
+  return has_fields
```
dbieber: lgtm.

fire/inspectutils.py
floscha: Fixed it with dabc6e6.

floscha:
No problem, it's already fixed ;)

dbieber:
Awesome. 👍

Issue #164, opened 2019-02-07
testFireKeywordArgs, testFireObjectWithDict, testHelpStringBuiltin, and testGetFullArgSpecFromBuiltin fail with Python 3.7


mcepl:
```
[   23s] =================================== FAILURES ===================================
[   23s] _________________________ FireTest.testFireKeywordArgs _________________________
[   23s] 
[   23s] self = <fire.fire_test.FireTest testMethod=testFireKeywordArgs>
[   23s] 
[   23s]     def testFireKeywordArgs(self):
[   23s]       self.assertEqual(
[   23s]           fire.Fire(
[   23s]               tc.Kwargs,
[   23s]               command=['props', '--name', 'David', '--age', '24']),
[   23s]           {'name': 'David', 'age': 24})
[   23s]       # Run this test both with a list command and a string command.
[   23s]       self.assertEqual(
[   23s]           fire.Fire(
[   23s]               tc.Kwargs,
[   23s]               command=['props', '--message',
[   23s]                        '"This is a message it has -- in it"']),  # Quotes stripped
[   23s]           {'message': 'This is a message it has -- in it'})
[   23s]       self.assertEqual(
[   23s]           fire.Fire(
[   23s]               tc.Kwargs,
[   23s]               command=['props', '--message',
[   23s]                        'This is a message it has -- in it']),
[   23s]           {'message': 'This is a message it has -- in it'})
[   23s]       self.assertEqual(
[   23s]           fire.Fire(
[   23s]               tc.Kwargs,
[   23s]               command='props --message "This is a message it has -- in it"'),
[   23s]           {'message': 'This is a message it has -- in it'})
[   23s]       self.assertEqual(
[   23s]           fire.Fire(tc.Kwargs,
[   23s]                     command=['upper', '--alpha', 'A', '--beta', 'B']),
[   23s]           'ALPHA BETA')
[   23s]       self.assertEqual(
[   23s]           fire.Fire(
[   23s]               tc.Kwargs,
[   23s] >             command=['upper', '--alpha', 'A', '--beta', 'B', '-', 'lower']),
[   23s]           'alpha beta')
[   23s] 
[   23s] fire/fire_test.py:240: 
[   23s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[   23s] 
[   23s] component = <class 'fire.test_components.Kwargs'>
[...]

PR #166, opened 2019-02-22
Fix issue #164

cshung:
This PR is meant for fixing https://github.com/google/python-fire/issues/164

After careful analysis, I believe the problem is that `inspect.getargspec()` is showing different behavior. In particular, the function throws an error when a built-in is given in Python 3.6 but it doesn't in Python 3.7.

The `testHelpStringBuiltin` and `testGetFullArgSpecFromBuiltin` cases are explicitly verifying that we use this base case so that it has `vars` and `kwargs`.

The `testFireKeywordArgs` and `testFireObjectWithDict` cases are more subtle than that - the specific return value of `inspect.getargspec()` is `FullArgSpec(args=['self'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})`, in particular, that `self` in `args` is causing the test cases to expect an extra parameter and leading to subsequent failures.

Once we move the `isbuiltin` test earlier, the behavioral difference are normalized and the test cases will all pass under both Python 3.6 and Python 3.7.

The official Python documentation does mention there is a change in 3.7, but enable inspection for built-in wasn't part of it. Search `getargspec` in the following page for more details.

https://docs.python.org/3/library/inspect.html

### DIFF
--- a/fire/inspectutils.py
+++ b/fire/inspectutils.py
@@ -74,10 +74,12 @@ def _GetArgSpecInfo(fn):
     if six.PY2 and hasattr(fn, '__init__'):
       fn = fn.__init__
   else:
-    # If the function is a bound method, we skip the `self` argument.
-    is_method = inspect.ismethod(fn)
-    skip_arg = is_method and fn.__self__ is not None
-
+    if inspect.ismethod(fn):
+      # If the function is a bound method, we skip the `self` argument.
+      skip_arg = fn.__self__ is not None
+    elif inspect.isbuiltin(fn):
+      # If the function is a bound builtin, we skip the `self` argument.
+      skip_arg = fn.__self__ is not None
   return fn, skip_arg
 
 
--- a/fire/helputils_test.py
+++ b/fire/helputils_test.py
@@ -88,7 +88,6 @@ class HelpUtilsTest(testutils.BaseTestCase):
     helpstring = helputils.HelpString('test'.upper)
     self.assertIn('Type:        builtin_function_or_method', helpstring)
     self.assertIn('String form: <built-in method upper of', helpstring)
-    self.assertIn('Usage:       [VARS ...] [--KWARGS ...]', helpstring)
 
   def testHelpStringIntType(self):
     helpstring = helputils.HelpString(int)
--- a/fire/inspectutils_test.py
+++ b/fire/inspectutils_test.py
@@ -56,8 +56,6 @@ class InspectUtilsTest(testutils.BaseTestCase):
     spec = inspectutils.GetFullArgSpec('test'.upper)
     self.assertEqual(spec.args, [])
     self.assertEqual(spec.defaults, ())
-    self.assertEqual(spec.varargs, 'vars')
-    self.assertEqual(spec.varkw, 'kwargs')
     self.assertEqual(spec.kwonlyargs, [])
     self.assertEqual(spec.kwonlydefaults, {})
     self.assertEqual(spec.annotations, {})


fire/inspectutils.py
```diff
@@ -100,8 +103,6 @@ def GetFullArgSpec(fn):
     # 1. If it's a builtin, it can take args.
```
dbieber: Thanks for finding the fix.

Can you update the comments to reflect your changes?

Also, are there any builtins that shouldn't return FullArgSpec(varargs='vars', varkw='kwargs')?

cshung: I moved the comments to reflect the fact that the built-ins are handled earlier.

To be compatible with Python 3.6 behavior, all built-ins that failed the `inspect.getargspec()` should return `FullArgSpec(varargs='vars', varkw='kwargs')`. By moving the code up, my assumption is that all built-ins would fail the call, but that's just my assumption.

dbieber: I did some investigation into isbuiltin, getargspec, and getfullargspec. Here's some of what I learned:

A list of Python 2 builtin functions is available here: https://docs.python.org/2/library/functions.html
There are 76 items in the list.
inspect.isbuiltin only returns True for 48 of them, returns False for 27 of them, and cannot be called on 1 of them (print).

For Python 3 the list of builtin functions is at https://docs.python.org/3/library/functions.html
There are 68 items on this list. inspect.isbuilting gives True for 41, False for 26, and cannot be run on 1 (breakpoint).

For many of these builtins, getargspec and getfullargspec successfully return the argspec of the builtin functions. For these, we want to use those argspecs. So, it sounds to me like the behavior is actually better in 3.7 than in 3.6, and we should update the tests to allow the improved behavior rather than updating the code to enforce the worse behavior.

cshung:
@dbieber 

Wonder if you get a chance to take a look at the new version?
I have updated the pull request so that it attempts to use the result of getargspec when it is available.

To do that, I have to remove a couple checks in the test case because they produce different results in Python 3.6 and 3.7, and also, I have to remove the self argument discovered by getargspec for built-ins.

Please take a look, thanks.

dbieber:
Glad you commented; I hadn't seen that you added a new version. I'll take a look.

fire/inspectutils.py
```diff
@@ -107,6 +107,9 @@ def GetFullArgSpec(fn):
   if skip_arg and args:
     args.pop(0)  # Remove 'self' or 'cls' from the list of arguments.
 
+  if args and args[0] == 'self':
```
dbieber: This should already be handled by 108.
If it wasn't, then skip_arg wasn't being set to True somewhere where it should have been set to True.

What example caused you to add this check?

cshung: The failing `testFireKeywordArgs` and `testFireObjectWithDict` cases did.

Maybe we should add the check 
`skip_arg = inspect.isbuiltin(fn) and fn.__self__ is not None` in `_GetArgSpecInfo` or even simpler, just check `fn.__self__ is not None` and ignore if it is a method or not?

Here was my earlier analysis:
> The `testFireKeywordArgs` and `testFireObjectWithDict` cases are more subtle than that - the specific return value of `inspect.getargspec()` is `FullArgSpec(args=['self'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})`, in particular, that `self` in `args` is causing the test cases to expect an extra parameter and leading to subsequent failures.

dbieber: I don't remember offhand why the ismethod check is needed, but I suspect there was a reason. (If we figure out that reason, let's add a comment saying what it is.)

Your idea of augmenting _GetArgSpecInfo so that skip_arg becomes True for the builtins that take self sgtm.

cshung: Just pushed a new version using the idea.

dbieber: Looks good to me 👍

dbieber:
This is merged (despite appearances in this PR) with 668007ae41391f5964870b4597e41493a936a11e. Thanks for making this fix.

Next, let's enable Python 3.7 in the .travis.yml file and update the setup.py to show that we support 3.7.

PR #170, opened 2019-03-28
Kafka producer and consumer example

kemalcanbora:
Kafka producer and consumer example for Google Fire


kemalcanbora:
I signed it!

PR #171, opened 2019-03-29
Kafka example for Python Fire

kemalcanbora:
This example about Kafka and python fire integration


dbieber:
Thanks for the submission. Good example :)
We're not going to accept this now, but will keep it in mind if/when we update documentation and examples in the future.

PR #192, opened 2019-09-01
add pypi version badge in README

williamfzc:
This useful badge helps developers know what the latest version number is directly.
Without this, we have to go to pypi.org to check the version number.


williamfzc:
@googlebot I signed it!)

PR #193, opened 2019-09-06
Exposes builtin functions from the standard library

### LINKED ISSUE #187: Cannot expose built-in functions from the standard library
First of all, thanks for the awesome library! It makes writing CLIs fun...because you don't have to!

I couldn't find a similar existing issue nor did I come across anything on Google or StackOverflow. I'm trying to expose the functions from the [math module](https://docs.python.org/3/library/math.html) in the standard library, but this error replicates with other standard library modules with built-in functions (like [operator](https://docs.python.org/3/library/operator.html)). I have the following in script.py:

```python
import math
import fire

if __name__ == "__main__":
    fire.Fire(math)
```

I can access constants in math fine:
```
$ python script.py pi
3.141592653589793
```

but not the functions (which happen to be built-in):
```
$ python script.py sin 3
  File "script.py", line 6, in <module>
    fire.Fire(math)
  File "/Users/bacon/miniconda/envs/fire/lib/python3.7/site-packages/fire/core.py", line 138, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/Users/bacon/miniconda/envs/fire/lib/python3.7/site-packages/fire/core.py", line 471, in _Fire
    target=component.__name__)
  File "/Users/bacon/miniconda/envs/fire/lib/python3.7/site-packages/fire/core.py", line 675, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
TypeError: sin() takes exactly one argument (0 given)
```

Other standard library modules like calendar and unicodedata work fine too, because their functions aren't built-ins. A work-around is to wrap all the functions that you want in user-defined functions, but that would a lot of boilerplate (and lack of boilerplate is why fire > argparse/anything else).

I'm on Python 3.7.3 and fire 0.2.1.

geoffbacon:
This PR exposes builtin functions from the standard library to fire and fixes #187.

Previously, certain builtin functions from the standard library (e.g. `math.sin`) would raise a TypeError complaining about getting one less required argument than they need. The original diagnosis was that these functions were written in C and did not expose their signature to`inspect.getfullargspec` or `inspect.signature`. However `math.sin` does expose its signature to these inspect functions:

```python
>>> inspect.getfullargspec(math.sin)
FullArgSpec(args=['x'], varargs=None, varkw=None, defaults=None, kwonlyargs=
[], kwonlydefaults=None, annotations={})
```

The problem was that `fire.inspectutils._GetArgSpecInfo` was incorrectly skipping the first argument of these builtins under the assumption that a non-null `__self__` attribute of a builtin always means that the function is a method and its first argument is `self`. However, for builtin functions in standard libraries, the `__self__` attribute is the module it came from. The proposed fix is to only skip the first argument of builtins if the `__self__` attribute isn't a module.

### DIFF
--- a/fire/inspectutils.py
+++ b/fire/inspectutils.py
@@ -19,6 +19,7 @@ from __future__ import division
 from __future__ import print_function
 
 import inspect
+import types
 from fire import docstrings
 
 import six
@@ -70,7 +71,7 @@ def _GetArgSpecInfo(fn):
   """
   skip_arg = False
   if inspect.isclass(fn):
-    # If the function is a class, we try to use it's init method.
+    # If the function is a class, we try to use its init method.
     skip_arg = True
     if six.PY2 and hasattr(fn, '__init__'):
       fn = fn.__init__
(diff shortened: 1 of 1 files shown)


dbieber:
Thanks for the fix! LGTM.
Travis is complaining about the indentation; we use 2 spaces.
Will squash and merge soon - prefer if you fix the indentation but if not I will fix in a follow up commit.

🎉

PR #194, opened 2019-09-14
Updated using-cli.md to reflect support for --help

pranavgupta1234:
Updated using-cli.md to reflect support for --help


dbieber:
Merged with modifications.

Issue #182, opened 2019-07-19
Support Bash Arrays


wahrwolf:
Hi there,
I love fire but I keep running in the same issue.

A lot of my scripts support a target directive so I can handover a list and it will iterate over the list.

If I combine that with Fire I need to write ugly python lists and can not use bash magic.
E.g. I need to write `python example.py count_lines '["file00", "file01", "file02"...]'` instead of something like 'python example.py count_lines "file{00..10}"` and I can not use any tab completion.

My current workaround is adding:
´´´
def count_lines(files):
  if isinstance(files, str):
    files = files.split()
  ...
```
One idea was to add this line with a special param like `--use-bash-arrays` to the Fire parser so I can run the script like: `python example.py count_lines "ffiles{00..}" -- --use-bash-arrays`

siddharths2710:
Hello @dbieber and @wahrwolf ,
I would like to contribute to this enhancement, and to this end, I'd like to discuss how to resolve this issue.
With the `--use-bash-arrays` parameter proposed by @wahrwolf , I have the following questions:


 - How can we infer the element type in the bash array? One solution that comes to mind is through another parameter which serves the purpose, ex. `--bash-array-elem-type=str`
 - Introducing a new command line parameter would mean to handle a similar named function argument in the Python code. For instance, ( although unlikely ) , the `count_lines` function may define a `use-bash-arrays` variable as a parameter in its signature. Any mechanism of handling this occurrence?
- A Bash array may also be specified like `( f1 f2 f3 ... )` . Should we accommodate such a declaration as well?

PR #200, opened 2019-10-08
Support Bash Arrays

tapaswenipathak:
Fixes https://github.com/google/python-fire/issues/182.


dbieber:
Thanks for the submission. We're not adding this feature at this time, so I'm going to close out the PR.

Issue #189, opened 2019-08-22
Docstring param description hidden when it contains a colon


bzamecnik:
Given a simple script:
```
import fire

def foo(name='World'):
    """
    Greets name.

    Arguments
    ---------
    name : str
        name, default: World
    """
    print('Hello %s' % name)

if __name__ == '__main__':
    fire.Fire(foo)
```

```
python foo.py -h
```

shows the help with the description of `--name` missing:

```
# ...
FLAGS
    --name=NAME
```

If the description does not contain a colon, like `name, default World` it works OK:

```
FLAGS
    --name=NAME
        name, default World
```

It should work even with a colon. I'd guess it's connected with parsing the `param_name : type` section which contains also a colon.

`fire==0.2.1`

dbieber:
This will require making changes to https://github.com/google/python-fire/blob/master/fire/docstrings.py

The `if` at https://github.com/google/python-fire/blob/master/fire/docstrings.py#L450, and particularly the `elif` at https://github.com/google/python-fire/blob/master/fire/docstrings.py#L457 are the first placed I would look. This elif controls the behavior when a colon is encountered in a numpy args section.

dbieber:
Fixed by #201 with b1f3f95d736e7e36126f2e6c457c52f995f587dc
Thanks Vincent!

PR #201, opened 2019-10-10
handle NumPy style docstring params containing a colon #189

vbarbaresi:
The problematic case was:

    name : str
        name, default: World

The second line wasn't treated as a description if it contained a colon
and ended up ignored

We parse line by line, and used to keep track only of the next line.
I added tracking of the previous line so that I can compare the indentations.
That allows to distinguish descriptions and a parameter definitions.

### DIFF
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -175,8 +175,9 @@ def parse(docstring):
 
   for index, line in enumerate(lines):
     has_next = index + 1 < lines_len
+    previous_line = lines[index - 1] if index > 0 else None
     next_line = lines[index + 1] if has_next else None
-    line_info = _create_line_info(line, next_line)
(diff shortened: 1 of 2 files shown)


dbieber:
Thanks for the contribution! 🙌 
It may be a bit before I get to review and accept it as I'm a bit bottlenecked by other work.

Thoughts from a quick first pass through the PR:
- logic looks good, OK to merge
- thanks for including a test
- please use consistent spacing and docstring formatting with the rest of the project (e.g. 2 spaces for indent, 4 space indent for continuations, 1-line summary at start of docstring / google-style docstrings)

dbieber:
We'll get this in shortly. Thanks again :)

dbieber:
Merged in b1f3f95d736e7e36126f2e6c457c52f995f587dc

Issue #190, opened 2019-08-22
Problem showing help when function is decorated


bzamecnik:
```python
import time
import fire

def decorator(f):
    def wrapper(*args, **kwds):
        return f(*args, **kwds)
    return wrapper

@decorator
def foo(name='World'):
    """
    Greets name.
    """
    print('Hello %s' % name)

if __name__ == '__main__':
    fire.Fire(foo)
```

produces:

```
# -h or --help
$ python bar.py -h
Traceback (most recent call last):
  File "bar.py", line 17, in <module>
    fire.Fire(foo)
  File "/Users/bzamecnik/.virtualenvs/rossum3/lib/python3.7/site-packages/fire/core.py", line 138, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/Users/bzamecnik/.virtualenvs/rossum3/lib/python3.7/site-packages/fire/core.py", line 471, in _Fire
    target=component.__name__)
  File "/Users/bzamecnik/.virtualenvs/rossum3/lib/python3.7/site-packages/fire/core.py", line 675, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "bar.py", line 6, in wrapper
    return f(*args, **kwds)
TypeError: foo() got an unexpected keyword argument 'h'
```

The same problem when the decorator uses `wraps()`.

```python
from functools import wraps

def timing(f):
    @wraps(f)
    def wrapper(*args, **kwds):
        return f(*args, **kwds)
    return wrapper
```

However, `python bar.py -- -h` shows the help but without the arguments:

```
NAME
    bar.py - Greets name.

SYNOPSIS
    bar.py <flags> [ARGS]...

DESCRIPTION
    Greets name.

POSITIONAL ARGUMENTS
    ARGS

FLAGS
    Flags are accepted.
```

When we remove the `@decorator` it works as expected:

```
NAME
    bar.py - Greets name.

SYNOPSIS
    bar.py <flags>

DESCRIPTION
    Greets name.

FLAGS
    --name=NAME
```

Versions: `fire=0.2.1, python 3.7.3`

## Workaround

A work around can be to move the actual code to a nested function that picks the arguments from the parent function. Then `fire` has not problem finding the arguments.

```python
from functools import wraps
import time
import fire

def timing(f):
    @wraps(f)
    def wrapper(*args, **kwds):
        return f(*args, **kwds)
    return wrapper

def foo(name='World'):
    """
    Greets name.
    """
[...]

Issue #159, opened 2018-12-28
Use inspect.signature()


robertvazan:
Use [inspect.signature()](https://docs.python.org/3.5/library/inspect.html#inspect.signature) if available, because `inspect.getfullargspec()` ignores annotations added by [functools.update_wrapper](https://docs.python.org/3/library/functools.html#functools.update_wrapper) to wrapper functions. Consequently any wrapper applied to functions in the module kills `--help` output and hurts interpretation of function arguments by making all parameters positional.

vbarbaresi:
I had a shot at this: https://github.com/vbarbaresi/python-fire/commit/38c154d9db2c7d2349d1eed8842ec6b93be42060


`inspect.getfullargspec()` is already kind of using `inspect.signature()` behind, just with legacy parameters:
(https://github.com/python/cpython/blob/3.8/Lib/inspect.py#L1090)
`skip_bound_arg=False` and `follow_wrapper_chains=False`

I overrode `inspect.getfullargspec()` with a method that reverses these parameters, and just removed the manual skipping of bound args:

```diff
-  if skip_arg and args:
-    args.pop(0)  # Remove 'self' or 'cls' from the list of arguments.
```
And it solves the `@lru_cache()` decorator case (#145)
But it doesn't solve #190, it looks like `-h` as argument without `--` is another code path

So I'm not really satisfied with it, but I can turn in into a PR if you think it's acceptable.


I should add that the lru_cache() decorator case is rather weird, I couldn't find a minimal reproduction case:

```python
import fire
from functools import update_wrapper


def decorator(maxsize=128, typed=False):
    def decorating_function(user_function):
        wrapper = _wrapper(user_function, maxsize, typed)
        return update_wrapper(wrapper, user_function)
    return decorating_function


def _wrapper(user_function, maxsize, typed):
    def wrapper(*args, **kwargs):
        result = user_function(*args, **kwargs)
        return result
    return wrapper

@decorator()
def foo(name):
    return('Hello %s' % name)

if __name__ == '__main__':
    fire.Fire()
```
This is as close as I could do from the actual lru_cache() decorator, but it doesn't trigger the bug.
So I had to really use functools.lru_cache in the tests..

dbieber:
Resolved by #203 (though looks likely to remain broken in Python 3.4)

PR #203, opened 2019-10-11
use a custom getfullargspec that follows wrapped chains #159

vbarbaresi:
using complex decorators like @functools.lru_cache()
was breaking positional parameters handling

We used `inspect.getfullargspec()`, that ignores `__wrapped__` attributes for legacy reasons
It is based internally on `inspect.signature()` which allows following the chain of
`__wrapped__` attributes

I took the code from getfullargspec() and changed it to use
skip_bound_arg=True and follow_wrapper_chains=True

This fixes the lru_cache decorator usage and allows to remove
a line that was getting rid of the bound args.

This applies only to Python3, Python2 wasn't affected by this issue

### DIFF
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -193,10 +193,12 @@ class CoreTest(testutils.BaseTestCase):
     )
 
   def testLruCacheDecoratorBoundArg(self):
-    self.assertEqual(core.Fire(tc.LruCacheDecoratedMethod, command=['lru_cache_in_class', 'foo']), 'foo')
+    self.assertEqual(core.Fire(tc.LruCacheDecoratedMethod,
+                               command=['lru_cache_in_class', 'foo']), 'foo')
 
   def testLruCacheDecorator(self):
     self.assertEqual(core.Fire(tc.lru_cache_decorated, command=['foo']), 'foo')
+
 
 if __name__ == '__main__':
   testutils.main()
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -192,5 +192,11 @@ class CoreTest(testutils.BaseTestCase):
         7,
     )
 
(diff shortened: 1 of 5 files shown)


vbarbaresi:
I wasn't sure at first but it fixes #190 actually
But it needs `@functools.wraps(f)` to be able to follow the chain. I'll add a test

dbieber:
Thanks a bundle for doing this. 

I always feel bad asking volunteers to do additional work, but if you're up for it, these are the change's we'll need to make:
- style stuff: 80 char line limit, 2 spaces for indent, 4 spaces for continuations, naming of custom_getfullargspec (I recommend GetFullArgSpecPy3), docstring for custom_getfullargspec (should be google-style, see other fns for examples)
- functools32 may provide an lru_cache that works in PY2 and PY3

Anything you don't get to I'll clean up when I accept the PR (which may be a bit, as I'm bottlenecked by other work at the moment.)

vbarbaresi:
no problem, I can handle the style stuff

for functools32, I'm not sure where to add development dependencies.
It looks like they are only embedded as `pip install foo` in the Travis config file?

dbieber:
Also as TEST_DEPENDENCIES in setup.py

vbarbaresi:
I just realized that functools32 is only made for Python2 so it would be a bit cumbersome to have some Py2 dependencies only.

The bug didn't occur on Py2 with `functools32.lru_cache()` anyway so it may not be necessary to really test it

dbieber:
OK, sgtm

cmin764:
Can't wait for this to be accepted!

dbieber:
I think we'll get this in shortly.

dbieber:
Merged in deb33cc34f94cad86a875a9993864ac0acf93828

Issue #142, opened 2018-09-29
Does python-fire have some method to do unified exception handling?


hujianxin:

kinow:
Came here looking for the same (as I knew [click had it](http://click.palletsprojects.com/en/5.x/exceptions/)). Couldn't find anything in docs/examples, so I guess it leaves that part to the user?

dbieber:
That's correct.

If there's an error caused by the user's usage (e.g. they try to access a command that doesn't exist, or don't provide required arguments for a command), then internally Fire raises a `FireError` with details about that error. Fire catches its own errors though, and will raise a `FireExit` (a subclass of `SystemExit` that contains information about the error) with status code 2, indicating a user error.

If there's a Python error in the CLI, then Fire doesn't catch it; it simply surfaces to the user as a Python error. You can certainly catch errors raised by CLI (both ordinary Python errors as well as `FireExit`s) if you want custom error handling code.

dbieber:
Here's the real answer to your question:

Yes, Fire catches `fire.core.FireErrors` and displays them succinctly alongside the usage information for the component at the time of the error.

So e.g. you can do something like

```
import fire, math
def sqrt(x):
  if x < 0:
    raise fire.core.FireError('x must be >= 0')
  return math.sqrt(x)
fire.Fire(sqrt)
```

Then when you use it you'll see:
```
> cli -2
ERROR: x must be >= 0
Usage: mathex.py X

For detailed information on this command, run:
  mathex.py --help
```

Sorry it's not in the docs yet!

dbieber:
Answered in previous comment; closing!

PR #215, opened 2020-01-11
Use "dir()" to get available methods of object

### LINKED ISSUE #149: Using __getattr__ breaks python-fire
This (admittedly hacky) metaprogramming way of implementing missing methods on the fly breaks python-fire.

Ref: https://stackoverflow.com/a/6955825/1096370

    import fire

    class Calculator(object):
        """A simple calculator class."""

        def double(self, number):
            return 2 * number

        def __getattr__(self, name):
            def _missing(*args, **kwargs):
                print("A missing method was called.")
                print("The object was %r, the method was %r. " % (self, name))
                print("It was called with %r and %r as arguments" % (args, kwargs))
            return _missing

    if __name__ == '__main__':
        fire.Fire(Calculator)

This "ghost methods" works normally when called from other python module.
When trying to call it with fire I get this:

    python -m calculator double 10

      File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
        "__main__", mod_spec)
      File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "/users/dev/project/calculator.py", line 19, in <module>
        fire.Fire(Calculator)
      File "/users/dev/project/venv/lib/python3.7/site-packages/fire/core.py", line 127, in Fire
        component_trace = _Fire(component, args, context, name)
      File "/users/dev/project/venv/lib/python3.7/site-packages/fire/core.py", line 437, in _Fire
        component, remaining_args)
      File "/users/dev/project/venv/lib/python3.7/site-packages/fire/core.py", line 513, in _GetMember
        members = dict(inspect.getmembers(component))
      File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 330, in getmembers
        for base in object.__bases__:
    TypeError: 'function' object is not iterable

meshde:
* Fix #149 
* Use `dir()` instead of `inspect.getmembers()` to get available methods of given object.
* Refer #149 for explanation.

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -632,7 +632,7 @@ def _GetMember(component, args):
   Raises:
     FireError: If we cannot consume an argument to get a member.
   """
-  members = dict(inspect.getmembers(component))
+  members = dir(component)
   arg = args[0]
   arg_names = [
       arg,
@@ -641,7 +641,7 @@ def _GetMember(component, args):
 
   for arg_name in arg_names:
     if arg_name in members:
-      return members[arg_name], [arg], args[1:]
+      return getattr(component, arg_name), [arg], args[1:]
 
   raise FireError('Could not consume arg:', arg)
 
--- a/fire/fire_test.py
+++ b/fire/fire_test.py
@@ -703,6 +703,11 @@ class FireTest(testutils.BaseTestCase):
     with self.assertRaisesFireExit(2):
       fire.Fire(tc.InstanceVars, command=['--arg1=a1', '--arg2=a2', '-', 'jog'])
 
+  def testClassWithDefaultMethod(self):
+    self.assertEqual(
+        fire.Fire(tc.DefaultMethod, command=['double', '10']), 20
+    )
+
 
 if __name__ == '__main__':
   testutils.main()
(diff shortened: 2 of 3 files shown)


dbieber:
Thanks for the PR!
LGTM at a first read.
Looks like the remaining travis failure is unrelated to your PR and you can ignore it.

dbieber:
For reference, here is the definition of getmembers:

https://github.com/python/cpython/blob/9dbf5d3bc2033940cdca35440cf08814544f81e4/Lib/inspect.py#L316-358.

It's gotten more complicated than it used to be in order to handle some edge cases.

dbieber:
Thanks for the contribution!

dbieber:
In addition to fixing #142, it also prevents properties from being evaluated unnecessarily. This prevents unintended side-effects (if evaluating a property has a side-effect) and fixes the case where evaluating a property would have raised an error.

PR #218, opened 2020-01-27
fix typos in guide

davidcaron:
- `{name: David}` -> `{name:David}` (no space)
- reference to last example should be the third example ?

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -693,28 +693,28 @@ The complete set of flags available is shown below, in the reference section.
 
 ### Reference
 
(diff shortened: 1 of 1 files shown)


davidcaron:
@googlebot I signed it!

dbieber:
Good catch on the last example. Should be either `'{name: David}'` or `{name:David}`.

The reference to the last example was as intended though. "name" and "David" are "bare-words", and are treated as strings (even though they don't have quotes around them) in the dict.

davidcaron:
Got it, I reverted that change.

PR #223, opened 2020-02-18
Support printing classes with overridden str method as SimpleGroups #197

### LINKED ISSUE #197: Help text uses class docstring instead of repr(object) in fire>=0.2.0 for objects in dict
dict with string values produces this CLI.

```
> python -c "import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key1: value1
key2: value2
```

dict with object values produced a very simlarly CLI in fire==0.1.3.

```
key1: Value(name='value1')
key2: Value(name='value2')
```

Since fire>=0.2.0 this is not the case anymore...

```
> python value.py
NAME
    value.py

SYNOPSIS
    value.py GROUP

GROUPS
    GROUP is one of the following:

     key1
       Docstring

     key2
       Docstring
```

Could you restore the old behavior?

**value.py**

```
class Value:
    """Docstring"""
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)

import fire
fire.Fire({'key1': Value('value1'), 'key2': Value('value2')})
```

jaredtrog:
Fixes #197 

Tested with the following classes:
```python
class Value(object):
    """Original question"""
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)


class NewValue:
    """As expressed in comments"""
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)
    def __str__(self):
        return repr(self)
```

### Python 2.7
```bash
% python --version
Python 2.7.16
% python -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key2: value2
key1: value1
```

### Python 3.7
```bash
% python3.7 --version                                                                                                     
Python 3.7.6
% python3.7 -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key1: value1
key2: value2
```

### Python 3.8
```bash
% python3.8 --version
Python 3.8.1
% python3.8 -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key1: value1
key2: value2
```

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -319,7 +319,8 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
   if verbose:
     return True
   # Python 3.7.6 evaluates type(absolute_import) == __future__._Feature while
-  # Python 2.7.16 (Base macOS Catalina) evaluates type(absolute_import) == <type 'instance'>
+  # Python 2.7.16 (Base macOS Catalina) evaluates
+  # type(absolute_import) == <type 'instance'>
   if isinstance(member, type(absolute_import)) and six.PY34:
     return False
   if inspect.ismodule(member) and member is six:
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -318,7 +318,9 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
     return False
   if verbose:
     return True
-  if isinstance(member, type(absolute_import)):
+  # Python 3.7.6 evaluates type(absolute_import) == __future__._Feature while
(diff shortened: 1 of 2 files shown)


jaredtrog:
@googlebot I fixed it.

jaredtrog:
Closing and reopening to correct the git commit author history.

PR #224, opened 2020-02-18
Support printing classes with overridden str method as SimpleGroups #197

### LINKED ISSUE #197: Help text uses class docstring instead of repr(object) in fire>=0.2.0 for objects in dict

jaredtrog:
Fixes #197 

Tested with the following classes:
```python
class Value(object):
    """Original question"""
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)


class NewValue:
    """As expressed in comments"""
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)
    def __str__(self):
        return repr(self)
```

### Python 2.7
```bash
% python --version
Python 2.7.16
% python -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key2: value2
key1: value1
```

### Python 3.7
```bash
% python3.7 --version                                                                                                     
Python 3.7.6
% python3.7 -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key1: value1
key2: value2
```

### Python 3.8
```bash
% python3.8 --version
Python 3.8.1
% python3.8 -c "import sys; sys.path.append('../python-fire'); import fire; fire.Fire({'key1': 'value1', 'key2': 'value2'})"
key1: value1
key2: value2
```

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -318,7 +318,10 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
     return False
   if verbose:
     return True
-  if isinstance(member, type(absolute_import)):
+  # Python 3.7.6 evaluates type(absolute_import) == __future__._Feature while
+  # Python 2.7.16 (Base macOS Catalina) evaluates
+  # type(absolute_import) == <type 'instance'>
+  if isinstance(member, type(absolute_import)) and six.PY34:
     return False
   if inspect.ismodule(member) and member is six:
     # TODO(dbieber): Determine more generally which modules to hide.
--- a/fire/value_types.py
+++ b/fire/value_types.py
@@ -37,7 +37,8 @@ def IsCommand(component):
 
 
 def IsValue(component):
-  return isinstance(component, VALUE_TYPES)
+  return isinstance(component, VALUE_TYPES) or (
+      hasattr(component, '__str__') and inspect.ismethod(component.__str__))
 
 
 def IsSimpleGroup(component):


dbieber:
Can you explain the inclusion of " and six.PY34" in the absolute_import check?

dbieber:
We're going to incorporate the main change here:
Objects with custom __str__ methods will be treated as Values, which means that a dict containing such objects will be serialized in one-line-per-key mode, rather than shown as a help screen.

I think this will usually provide a better experience. However, one situation that it makes worse is this:
If a user has defined a multiline __str__ method for an object (e.g. in order to provide a custom help/output experience for users of that object) and then they nest that kind of object inside a dict, now the output for that dict will be messy.
This situation seems relatively uncommon though, so the change seems worthwhile.

fire/value_types.py
```diff
@@ -37,7 +37,8 @@ def IsCommand(component):
 
 
 def IsValue(component):
-  return isinstance(component, VALUE_TYPES)
+  return isinstance(component, VALUE_TYPES) or (
+      hasattr(component, '__str__') and inspect.ismethod(component.__str__))
```
dbieber: I think we're going to do this a little differently, using the same logic as in core.py:_PrintResult to determine what counts has having a custom __str__ method.

No need to make any changes though; we'll accept your PR and then add a follow-on commit with the change.

fire/completion.py
```diff
@@ -318,7 +318,10 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
     return False
   if verbose:
     return True
-  if isinstance(member, type(absolute_import)):
+  # Python 3.7.6 evaluates type(absolute_import) == __future__._Feature while
+  # Python 2.7.16 (Base macOS Catalina) evaluates
+  # type(absolute_import) == <type 'instance'>
+  if isinstance(member, type(absolute_import)) and six.PY34:
```
dbieber: Could you explain why this is only needed in PY34?
Is the existing behavior harmful in PY27?

jaredtrog: Hey David, thanks for the review!

I found that the following test used the old behavior and printed as expected on 3.7 but 2.7 still gave the help text back.

```
import sys
sys.path.append('../python-fire')
import fire


class Value(object):
    """Original question"""

    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)


class NewValue:
    """As expressed in comments"""

    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return "%s(name=%r)" % (self.__class__.__name__, self.name)

    def __str__(self):
        return repr(self)

# Should print help
fire.Fire({'key1': Value('value1'), 'key2': Value('value2')})
# Should print SimpleGroup
fire.Fire({'key1': NewValue('value1'), 'key2': NewValue('value2')})
```

At this point I had implemented the modification to `IsValue` in `value_types`:

```
def IsValue(component):
  return isinstance(component, VALUE_TYPES) or (
      hasattr(component, '__str__') and inspect.ismethod(component.__str__))
```


I went to verify the criteria for a SimpleGroup as you outlined in the recommended approach; `core.py` -> `value_types.py` -> `IsValue(component)`. The criteria is that the `result` is a dictionary and that `value_types.IsSimpleGroup(result)` return a truthy value: (I know you know, just brain dumping what I was looking at.)

https://github.com/google/python-fire/blob/04e2434001e809bdf05e6f47d71452d1e8614b78/fire/core.py#L261-L262

The value is obviously a dictionary so I checked the return of `value_types.IsSimpleGroup(result)` and found that it was returning `True` and I was dropping into line 262 attempting to print the dictionary as string but... it was returning an empty dictionary; but only on Python 2.7.

So I took a look at `_DictAsString`, specifically at this section:

https://github.com/google/python-fire/blob/04e2434001e809bdf05e6f47d71452d1e8614b78/fire/core.py#L315-L323

After a bit of debug I found that the `MemberVisible` test was returning `False`. This seemed odd considering there was different behavior on Python 3.7 so I kept going.

I took a look at `completion.MemberVisible` and iterated through the conditional list to find which was was returning `False` and found it was the following statement:

https://github.com/google/python-fire/blob/04e2434001e809bdf05e6f47d71452d1e8614b78/fire/completion.py#L321-L322

Knowing that the test results were different between 2.7 and 3.7, I popped the statements in the repl to see the outputs.

```
Python 2.7.16 (default, Dec 13 2019, 18:00:32) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.32.4) (-macos10.15-objc-s on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import absolute_import
>>> type(absolute_import)
<type 'instance'>
```

vs

```
Python 3.7.6 (default, Dec 30 2019, 19:38:26)
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import absolute_import
>>> type(absolute_import)
<class '__future__._Feature'>
```

I have not discovered exactly why the 2.7 `__future__.absolute_import` returns an instance and 3.5+ `__future__.absolute_import` returns a class. It seems there may be a difference in `import` between 2.7 and 3 where 2.7 returns an instance of the module and 3.7 returns the actual class reference.

The other possiblity is that I've missed something obvious and it's none of the above :).

Either way, the functionality of this if statement seems to be to hide instances of `__future__` probably to keep the noise down. Given what I understand of the import behavior between 2.7 and 3.4+, this result will only be achieved if the version of python is 3.4+ as currently implemented. The `six` module provides the builtin to test the version of python and I see `six.PY27` is used elsewhere so I chose to lock this conditional to 3.4+.

I'm sure there are other approaches. This approach seemed to require the fewest changes to the code while retaining functionality. It is quite possible that I've missed the intent of the if or some other functionality it provides.

fire/value_types.py
jaredtrog: Sounds good - thanks!

fire/completion.py
dbieber: Thanks for the explanation.

> the functionality of this if statement seems to be to hide instances of __future__ probably to keep the noise down

Yes that's right.

If I'm understanding correctly, in Python 2.7 the check for type(absolute_import) is preventing all things of type <type 'instance'> from being shown, and so it's being too aggressive in what it's hiding.

jaredtrog: > If I'm understanding correctly, in Python 2.7 the check for type(absolute_import) is preventing all things of type <type 'instance'> from being shown, and so it's being too aggressive in what it's hiding.

Yes, that seems to be correct.

dbieber:
Merged in 7d87eb2

dbieber:
Also incorporated your advice about absolute_import in c8ea506b067ac13465f0331d54afce35e495da3f

PR #227, opened 2020-02-25
Fix encoding errors

cauefcr:
Hi, i was using this library, and it had difficulty dealing with some utf-8 outputs, so i added the necessary changes to make it work.


cauefcr:
@googlebot I signed it

dbieber:
Your CL has 45 files changed. And a bunch of code commented out. Could you clean it up or open a new PR with a clean version, perhaps?
Then I could more easily see what changes were needed to get the encoding right.

dbieber:
If you have simple example where Fire failed that would be useful too.

Issue #261, opened 2020-05-21
Multi-line parameter descriptions in Google-style docstrings


lukedeo:
Hey there! Quick question regarding the behavior of multi-line parameter descriptions in Google-style docstrings. 

Consider the following code:

```python
# main.py
def foo(bar):
    """
    This does a foo on a bar.

    Args:
        bar (str): Though a simple parameter, this parameter will need a lot of
            verbiage to describe, requiring a second line.
    """
    print(f"Hi, {bar}!")

if __name__ == "__main__":
    import fire

    fire.Fire(foo)
```

When running `main.py --help`, the resulting `man` is:

```
NAME
    main.py - This does a foo on a bar.

SYNOPSIS
    main.py BAR

DESCRIPTION
    This does a foo on a bar.

POSITIONAL ARGUMENTS
    BAR
        Though a simple parameter, this parameter will need a lot of

NOTES
    You can also use flags syntax for POSITIONAL ARGUMENTS
```
Notice that we are losing the following line here.

I'd expect the newline-containing description to be un-newlined (definitely not a word)

```
NAME
    main.py - This does a foo on a bar.

SYNOPSIS
    main.py BAR

DESCRIPTION
    This does a foo on a bar.

POSITIONAL ARGUMENTS
    BAR
        Though a simple parameter, this parameter will need a lot of verbiage to describe, requiring a second line.

NOTES
    You can also use flags syntax for POSITIONAL ARGUMENTS
```

Or maybe included with the newline, but in some way retaining the following line. Any plans to support this, or (likely) am I missing something?

dbieber:
I actually thought we supported this already - I'll mark it as a bug. Yes, we would like for this to work.

lukedeo:
Thanks @dbieber!

MichaelCG8:
I've seen the same using Numpy style docstrings, I'll find an example and add it later. I'll look into this tonight and see if it's something I can help with.

MichaelCG8:
I've experimented and found the following.

    def foo(bar):
    """
    This does a foo on a bar.

    Args:
        bar (str): Though a simple parameter, this parameter will need a lot of
            verbiage to describe, requiring a second line.
    """
    print(f"Hi, {bar}!")

    def foo2(bar):
      """
      This does a foo on a bar.

      Parameters
      ----------
      bar : str
          Though a simple parameter, this parameter will need a lot of
          verbiage to describe, requiring a second line.
      """

    def foo3(bar):
      """
[...]

PR #262, opened 2020-05-23
issue-261 Fixed missing parts of argument descriptions

MichaelCG8:
issue-261 Fixed missing parts of argument descriptions for Google and Numpy style docstrings.

### DIFF
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -295,9 +295,8 @@ def _is_arg_name(name):
   """Returns whether name is a valid arg name.
 
   This is used to prevent multiple words (plaintext) from being misinterpreted
-  as an argument name. So if ":" appears in the middle of a line in a docstring,
-  we don't accidentally interpret the first half of that line as a single arg
-  name.
+  as an argument name. Any line that doesn't match the pattern for a valid
+  argument is treated as not being an argument.
 
   Args:
     name: The name of the potential arg.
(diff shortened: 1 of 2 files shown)


dbieber:
Belated thanks for this PR!

If I'm understanding correctly:

There are two behavior changes within:
1. We now use `arg_pattern = r'^[a-zA-Z_]\w*$'` for determining valid argument names.
2. We now set `state.current_arg = arg` when we encounter a Google arg-and-type.

Change 1. doesn't seem so important -- it prevents the parser from treated some invalid arg names as args, but as I understand it this increases strictness hasn't come up in practice.

Change 2. is important; it fixes a bug where previously we were losing some information about multiline arg descriptions.

Can you confirm my understanding is correct?

Thanks.

Thanks also for the test cases 👍

MichaelCG8:
With similar belatedness, thank you for reviewing!

**Change 1**
The old code would treat a line as an arg name if it is one word long and doesn't contain a `:`. The case that fails is any multiline description where the last line consists of one word only, as in `foo3` https://github.com/google/python-fire/issues/261#issuecomment-632889757

This could have been fixed  by also checking for a `.` in the string, so that any text ended with a period would be parsed as the plaintext description. The regex is more general and fixes the behaviour for any string that isn't a valid argument pattern, for example if the docstring ends with a reference to a type by surrounding it in back ticks.

The case where there is a single word followed by a period is something I've hit a few times if I reach my line length limit and wrap the last word onto a new  line.

**Change 2**
Spot on.

dbieber:
Closed with c39de6ae333fdaf7d3371ab7cb213cb28703120d

PR #265, opened 2020-06-09
Update: unittest extension for python 3.x

AIweimd:
Hi,
  The mock object library has been integrated into the unittest module in python3.x, so I try to add extention like this.

`try:`

    import mock  # python 2.x

`except ModuleNotFoundError:`

    from unittest import mock  # python 3.x

Thanks for watching.

### DIFF
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -22,7 +22,11 @@ from fire import core
 from fire import test_components as tc
 from fire import testutils
 from fire import trace
-import mock
+
+try:
+  import mock  # python 2.x
+except ModuleNotFoundError:
+  from unittest import mock  # python 3.x
 
 import six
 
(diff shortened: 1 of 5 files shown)


dbieber:
Thanks for the PR!

If I understand correctly, `mock` as a package from pypi continues to work in Python 3, and this change just allows us to drop the dependency on mock for testing purposes for Python 3.

Can you confirm I'm understanding correctly?

Could you also sign the CLA, please?

Borda (approved):

a-detiste:
The try/except should be the other wat around ( unittest.mock is newer than mock nowadays ),
I can provide a PR

dbieber:
We've moved entirely to unittest.mock https://github.com/google/python-fire/pull/545

PR #275, opened 2020-07-27
Issue 259: Add :key: parameters to help string

### LINKED ISSUE #259: :key: parameter documentation doesn't correctly display in help signature
Consider the following application:

```
import fire

def test(a, b, **test):
  """
  Returns the operation specified on a and b.
  
  :param a: The first parameter
  :param b: The second parameter
  
  :key operation: What to do with these parameters
  """
  op = test.get('operation', sum)
  return op([a,b])
  
fire.Fire(test)
```

Since `:key:` is recognized as sphinx documentation for keyword arguments, I would expect Fire to show it similar to :param: docstrings. However, the printout ends up looking like this:

```
POSITIONAL ARGUMENTS
    A
        The first parameter
    B
        The second parameter

        :key operation: What to do with these parameters
```

Where the key is not displayed as a separate argument. Compare that to the PyCharm help signature:
image

amy-lei:
Resolves #259 

Add `:key:` parameters under flags as so
```
FLAGS
    --arg=ARG
    The following flags are also accepted.
    --key
        Description of the `key` kwarg from the docstring
    Additional undocumented flags may also be accepted.
```

### DIFF
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -77,6 +77,11 @@ class ArgInfo(
 ArgInfo.__new__.__defaults__ = (None,) * len(ArgInfo._fields)
 
 
+class KwargInfo(ArgInfo):
+  pass
+KwargInfo.__new__.__defaults__ = (None,) * len(KwargInfo._fields)
+
+
 class Namespace(dict):
   """A dict with attribute (dot-notation) access enabled."""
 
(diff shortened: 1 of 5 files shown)


amy-lei:
I opened a draft PR for now to get feedback on the approach since it is my first time contributing to python-fire! Other ideas I've considered include adding a field to `DocstringInfo` for kwargs or adding a field to ArgInfo to differentiate between positional args and kwargs

dbieber:
Thanks for the PR! I haven't had a chance to look in detail yet. Will let you know once I do.
🙏

dbieber:
Merged with aa18534de3e1f44ed3dd525162568605a13e839a

Thanks for the contribution!

dbieber:
Merged in

Issue #245, opened 2020-04-19
AttributeError when calling a method which returns a BeautifulSoup object


gab50000:
This is a small example to reproduce the error:

```
import bs4
import fire


class Test:
    def get_soup(self):
        soup = bs4.BeautifulSoup("<h1>hi</h1>")
        return soup


fire.Fire(Test)
```

If I run this file via

```
python test.py get_soup
```

I get error "AttributeError: 'NoneType' object has no attribute 'get'".

If I run

```
python test.py get_soup -- --interactive
```

The variable result does not contain the expected soup object, but instead "<bound method Test.get_soup of <__main__.Test object at 0x7f80840e5460>>"

WaizungTaam:
The error log is
```
...\fire\core.py", line 775, in _ParseArgs
    accepts_positional_args = metadata.get(decorators.ACCEPTS_POSITIONAL_ARGS)
AttributeError: 'NoneType' object has no attribute 'get'
```

From https://github.com/google/python-fire/blob/master/fire/decorators.py#L87
```python
def GetMetadata(fn):
  # Class __init__ functions and object __call__ functions require flag style
  # arguments. Other methods and functions may accept positional args.
  default = {
      ACCEPTS_POSITIONAL_ARGS: inspect.isroutine(fn),
  }
  return getattr(fn, FIRE_METADATA, default)
```
`GetMetadata` uses `getattr` to retrieve the fire metadata from component `fn`. In this case, `fn` is a `bs4.BeautifulSoup`.

For `bs4.BeautifulSoup` ([bs4/\_\_init\_\_.py#L68](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/__init__.py#L68)), it inherits from class `bs4.Tag`, which overrides `__getattr__` ([bs4/element.py#L1433](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py#L1433)).
```python
def __getattr__(self, tag):
        """Calling tag.subtag is the same as calling tag.find(name="subtag")"""
        ...
```
So `getattr(fn, FIRE_METADATA, default)` for a `BeautifulSoup` instance always return `None` for the non-existing subtag, and calling `metadata.get` later raises the above exception.

A quick fix is to check if the `getattr` result is a `dict` with key `ACCEPTS_POSITIONAL_ARGS`, and return `default` if it is not.

dbieber:
[...]

PR #281, opened 2020-08-23
Beautiful Soup issue fix

roopeshvs:
Beautiful Soup object receiving NoneType error in getattr().
Fixed as suggested in #245


dbieber:
Merged (with modifications) in bc43dca

Thanks for your contribution!

Issue #272, opened 2020-07-14
complex objects in properties not possible?


michaelaye:
Using fire version 0.3.1, I have this simple code failing for me:

```python
import numpy as np
import pandas as pd

class MyClass:
    @property
    def data(self):
        return pd.DataFrame(np.random.rand(10, 3))

if __name__ == "__main__":
    fire.Fire(MyClass)

```

Getting these errors, when calling without options to see the help text:
```python
Traceback (most recent call last):
  File "pet_weights.py", line 69, in <module>
    fire.Fire(MyClass)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/core.py", line 164, in Fire
    _PrintResult(component_trace, verbose=component_trace.verbose)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/core.py", line 267, in _PrintResult
    result, trace=component_trace, verbose=verbose)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/helptext.py", line 63, in HelpText
    actions_grouped_by_kind = _GetActionsGroupedByKind(component, verbose=verbose)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/helptext.py", line 332, in _GetActionsGroupedByKind
    members = completion.VisibleMembers(component, verbose=verbose)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/completion.py", line 366, in VisibleMembers
    (member_name, member) for member_name, member in members
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/completion.py", line 368, in <listcomp>
    verbose=verbose)
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/fire/completion.py", line 311, in MemberVisible
    if member in (absolute_import, division, print_function):
  File "/home/maye/miniconda3/envs/py37/lib/python3.7/site-packages/pandas/core/generic.py", line 1479, in __nonzero__
    f"The truth value of a {type(self).__name__} is ambiguous. "
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

bradleydamato:
@dbieber can we just use `type(member)` instead of `member` here? It works on my local when I make that change. Happy to PR if you think it will work.

dbieber:
[...]

PR #290, opened 2020-10-02
Converting "member" to "type(member)" on line 311 of completion.py to fix issue 272

bradleydamato:
Passes: 
`completion_test.py`
`core_test.py`
`main_test.py`

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -308,7 +308,7 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
     return False
   if verbose:
     return True
-  if member in (absolute_import, division, print_function):
+  if type(member) in (absolute_import, division, print_function):
     return False
   if isinstance(member, type(absolute_import)) and six.PY34:
     return False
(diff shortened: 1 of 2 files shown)


bradleydamato:
@dbieber here is the PR for Issue 272 https://github.com/google/python-fire/issues/272

fire/completion.py
```diff
@@ -308,7 +308,7 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
     return False
   if verbose:
     return True
-  if member in (absolute_import, division, print_function):
+  if type(member) in (absolute_import, division, print_function):
```
dbieber, line 311: I don't think this is quite the check we want.

In Python 3 type(absolute_import) is __future__._Feature, which is fine.
However in Python 2 type(absolute_import) is instance which is too broad.

Instead we can do `if (member is absolute_import or member is division or member is print_function)`

bradleydamato: Sounds good @dbieber I'll make the change and push again

bradleydamato:
Looks like the line is too long; third time is the charm :). BRB.

dbieber:
Looks good, thanks for the PR.

dbieber:
Merged in c3789064e2e161992e3c5a23eb0c47b66072707f

PR #296, opened 2020-11-04
Don't run pytype in Python 3.5.

rchen152:
The next release of pytype will drop support for running under Python 3.5: https://github.com/google/pytype/issues/677.


dbieber:
Thanks for the update!

dbieber:
Merged in a938ef9

PR #303, opened 2020-12-25
use of hasattr, fix potential bug risk

tusharnankani:


tusharnankani:
@googlebot I signed it!

PR #304, opened 2020-12-25
change in behaviour, optimize code

tusharnankani:


dbieber:
Thanks for the PR. 🙏 
We generally don’t accept changes to the console module directly in the fire repo though, in order to keep any diffs between it and its upstream source easy to reason about.

Is there a reason to prefer AssertionErrors over asserts?

tusharnankani:
> Is there a reason to prefer AssertionErrors over asserts?

Whenever the use of assert is detected, the enclosed code will be removed when compiling to optimized byte code.
The assert statement is generally for unit-testing or finding issues during the development process. This does not mean you will necessarily see the error. On the other hand, raising an AssertionError means that the error will be raised by the code and that any calling processes can deal with this error as needed.


References for the same^
https://stackoverflow.com/questions/48408692/raise-assertionerror-vs-assert-python
https://stackoverflow.com/questions/40182944/difference-between-raise-try-and-assert
https://github.com/plyara/plyara/issues/78


> We generally don’t accept changes to the console module directly in the fire repo though, in order to keep any diffs between it and its upstream source easy to reason about.

Apologies from my end, I wasn't aware of that. Let me know what should I do next?

tusharnankani:
@dbieber Any updates?

dbieber:
> Apologies from my end, I wasn't aware of that.
No worries.

> @dbieber Any updates?
I'll let you know about making changes to console if I figure it out in the new year. Otherwise, no action to take on your part for now.

Issue #293, opened 2020-10-22
string literals in lists can't start with numbers


thomasgilgenast:
Following the example from [the fire guide section on argument parsing](https://github.com/google/python-fire/blob/master/docs/guide.md#argument-parsing), it seems like when passing a list of strings to fire, none of the elements of the string can start with a number, unless two extra pairs of quotes are used (no single pair of quotes works, at least in bash).

```bash
$ echo "import fire; fire.Fire(lambda obj: type(obj).__name__)" > test.py
$ python test.py [hi]
list
$ python test.py [4hi]
str
$ python test.py '["4hi"]'
list
```

This is because `ast.parse()` (called in `_LiteralEval()`) chokes on arguments like `'[4hi]'`:

```
Python 3.7.4 (default, Sep 12 2019, 15:40:15)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from fire.parser import _LiteralEval
>>> _LiteralEval('[hi]')
['hi']
>>> _LiteralEval('[4hi]')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/fire/parser.py", line 96, in _LiteralEval
    root = ast.parse(value, mode='eval')
  File "/usr/local/lib/python3.7/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    [4hi]
       ^
SyntaxError: invalid syntax
>>> _LiteralEval('["4hi"]')
['4hi']
```

My understanding here is limited, but I think `ast.parse()` succeeds when the string does not start with a letter because it parses `[hi]` to a list containing only one element: the variable (not a string) `hi`. This variable `hi` in the AST is later replaced with the string literal `'hi'` by `_Replacement()`. Once the variable has been replaced by the string literal in the AST, the AST can be evaluated by `ast.literal_eval()`.

When an element in the list begins with a number (e.g., when the argument to `ast.parse()` is `'[4hi]'`), `ast.parse()` again interprets this as a list containing only one element: the variable `4hi`. This time, `4hi` is not a valid Python identifier (since variable names cannot start with numbers), so `ast.parse()` crashes.

[...]

PR #307, opened 2021-01-13
Fix #229, #293: Modify parser.py to re-insert quotes correctly

### LINKED ISSUE #229: All the list is convert into string if '-' is present
Hello,

Capture d’écran 2020-03-02 à 09 56 52

I reuse the introduction example code to illustrate.
When you give a list as arg, and into this list there is a string including this specific charactere '-', Fire turns all the list into one and unique string 😢. As you can see with the capture.

I'm looking into the fire code to see where the modification could be.

Thanks !

stephenvincent27:
Passes:
`parser_test.py`
`parser_fuzz_test.py (?)`
`main_test.py`

We are not Python programmers primarily, so do let us know if there are fragments of code that can be rewritten in Python to make it more readable. Other than that, this code conforms faithfully to the flowchart we sent over earlier.

### DIFF
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -73,7 +73,6 @@ def DefaultParseValue(value):
     # If _LiteralEval can't parse the value, treat it as a string.
     return value
 
-
 def _LiteralEval(value):
   """Parse value as a Python literal, or container of containers and literals.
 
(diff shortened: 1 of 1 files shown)


stephenvincent27:
@googlebot I signed it!

dbieber:
Thank you for the PR!
I will let you know once I have looked through it more closely.

You mentioned in our discussion on #229 that you had some test cases. Could you add those to this PR?

stephenvincent27:
Hello @dbieber! Really sorry for the super-long hiatus. I've added relevant tests (some of which we'd discussed earlier), feel free to suggest additions/modifications to them.

dbieber:
Hello again. Thanks for adding the tests!

PR #316, opened 2021-02-04
Fix #315 Change the _KeywordOnlyArguments filtering condition

### LINKED ISSUE #315: Optional and required flags are swapped in the "Usage" message
Running python script.py command

```
import fire


class Example:

    def command(self, *, arg1: bool, arg2: bool = True, arg3: bool = False):
        pass


if __name__ == '__main__':
    fire.Fire(Example)

```

Prints

```
ERROR: Missing required flags: {'arg1'}
Usage: download.py command <flags>
  optional flags:        --arg1
  required flags:        --arg2 | --arg3
```

The "ERROR" message correctly points that arg1 is required, but the "Usage" message is swapping required and optional arguments.

xavierfigueroav:
Fix #315 by changing the _KeywordOnlyArguments filtering condition.

The condition checked for the presence of the flag in the `kwonlydefaults` field of `FullArgSpec` to consider it as required, but that actually proved the opossite.

### DIFF
--- a/fire/helptext.py
+++ b/fire/helptext.py
@@ -700,7 +700,7 @@ def _GetCallableUsageItems(spec, metadata):
 
 def _KeywordOnlyArguments(spec, required=True):
   return (flag for flag in spec.kwonlyargs
-          if required == (flag in spec.kwonlydefaults))
+          if required != (flag in spec.kwonlydefaults))
 
 
 def _GetCallableAvailabilityLines(spec):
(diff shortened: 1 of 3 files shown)


HLA55:
Up

dbieber:
I'm going to merge this and then clean things up in a follow-up PR.

The change we'll need to make is `def identity2(self, *, alpha, beta='0'):` needs to be moved to test_components_py3, since this isn't valid syntax for python 2.

dbieber:
^Done in #389.

PR #318, opened 2021-02-06
Fix #317 Help text short args

### LINKED ISSUE #317: Feature request short arguments in help text
Saw short arguments were added in #141, which looks great!
 
I was wondering if it'd be possible to include the short args in help text too?

I was thinking something like, given:
```python
# foo.py
import fire

def func(foo=False, bar='bar', baz='dupe'):
    pass

if __name__ == '__main__':
    fire.Fire(func)
```
The output could change like this:
```diff
# python foo.py --help
 FLAGS
-    --foo=FOO
+    -f, --foo=FOO
         Default: False
     --bar=BAR
         Default: 'bar'
     --baz=BAZ
         Default: 'dupe'
```
I'd be happy to try and pr this myself if you think it's worth adding?

ConorSheehan1:
Closes #317

### DIFF
--- a/fire/helptext.py
+++ b/fire/helptext.py
@@ -258,7 +258,9 @@ def _ArgsAndFlagsSections(info, spec, metadata):
     for flag in flags:
       if isinstance(flag, docstrings.KwargInfo):
         if flag.name[0] in unique_short_flags:
-          flag_string = short_flag_string.format(name=flag.name, short_name=flag.name[0])
+          flag_string = short_flag_string.format(
+            name=flag.name, short_name=flag.name[0]
+          )
         else:
           flag_string = flag_string.format(name=flag.name)
 
--- a/fire/helptext.py
+++ b/fire/helptext.py
@@ -33,9 +33,9 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import collections
 import itertools
 import sys
-from collections import Counter
 
 from fire import completion
 from fire import custom_descriptions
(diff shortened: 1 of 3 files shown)


dbieber:
Thanks for the PR!

ConorSheehan1:
Hey @dbieber no worries, I use fire for a few projects so I'd love to contribute back!

Also, I noticed the github action on my fork was failing because it couldn't find python 3.4
https://github.com/ConorSheehan1/python-fire/runs/1842567364?check_suite_focus=true
```
Version 3.4 was not found in the local cache
Error: Version 3.4 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

Only way I could get the build working was by removing 3.4, but I didn't include that in the pr.
Happy to add it in though if you think it's needed?
 
Also checked locally that my branch still passes on 3.4.
py-3 4 10-fire-crop

dbieber:
I've removed 3.4 from the CI now. Would be nice to add it back in if we can do so properly, but it's certainly not a priority.

ConorSheehan1:
@dbieber cool, is there anything left I should change or is this ok to merge now?

dbieber:
I'll do some cleanup when I merge it but unfortunately I won't get to that today.

I'm also not sure why the CI tests didn't run on this PR...

ConorSheehan1:
Cool no rush. Not sure about the CI either tbh, can you trigger it manually? Or can I do something on my end?

dbieber:
Thanks for your patience. At long last, I've merged your PR and we'll include it in the next release.

PR #320, opened 2021-02-16
Fix #300 typed arguments

### LINKED ISSUE #300: [FEATURE REQUEST] typed arguments
If I write a typed variable argument, I am expecting an argument of a specific type.
Example:

```python
import fire

def test(some_str: str):
    return print(some_str)

if __name__ == "__main__":
    fire.Fire()
```

but 

```shell
python test.py test 7.30
7.3
```

So this feature will not break [argument parsing](https://github.com/google/python-fire/blob/master/docs/guide.md#argument-parsing). The typed argument should be enough to detect the real type.

loynoir:
# Goal
[x] Solve [issue300](https://github.com/google/python-fire/issues/300)
[x] Let python-fire respect function's type annotations, and cast arguments into right types.
[x] Solve [issue64](https://github.com/google/python-fire/issues/64)

# Status
1. If exists annotation, simple type (str, int, float) arguments are casted correctly now.
2. [iln-sk](https://github.com/iln-sk) add `datetime` and `pathlib` typed.
3. completion zsh compactible.

# Now support
```python
def hello(self, x: str, y: int, z: float):
    assert type(x) == str
    assert type(y) == int
    assert type(z) == float
```

# Not Support
Custom types in annotations.


loynoir:
@googlebot I signed it!

baryluk:
This would be super useful. Right now we need to specify a lot of decorators to provide similar functionality, because very often our arguments looks like integers, and they are converted to int automagically, which triggers code later to fail (i.e. because it tries to use regexp on it, or concat with other string, etc).

Another good feature would be ability to disable automatic type detection also for `kwargs`, if it is typed, i.e. `Dict[str, str]`.

PR #326, opened 2021-03-19
Update guide.md

Lix1993:
fix the example since `python example.py hello world` will rise an error.


Lix1993:
@googlebot I signed it!

Issue #333, opened 2021-04-23
Output of the example for grouping in docs is wrong


wchliao:
Hello.

In the Group Commands section (https://github.com/google/python-fire/blob/master/docs/guide.md#grouping-commands), the example code is

```python
class IngestionStage(object):

  def run(self):
    return 'Ingesting! Nom nom nom...'

class DigestionStage(object):

  def run(self, volume=1):
    return ' '.join(['Burp!'] * volume)

  def status(self):
    return 'Satiated.'

class Pipeline(object):

  def __init__(self):
    self.ingestion = IngestionStage()
    self.digestion = DigestionStage()

  def run(self):
    self.ingestion.run()
    self.digestion.run()
    return 'Pipeline complete'

if __name__ == '__main__':
  fire.Fire(Pipeline)
```

and according to the guide

```bash
$ python example.py run
Ingesting! Nom nom nom...
Burp!
```

However, the output is wrong. It should be

```bash
$ python example.py run
Pipeline complete
```

This can be easily fixed.
However, in order to demonstrate the pipeline functionality, I would suggest to change the example to

```python
class IngestionStage(object):

  def run(self):
    return 'Ingesting! Nom nom nom...'

class DigestionStage(object):

  def run(self, volume=1):
    return ' '.join(['Burp!'] * volume)

  def status(self):
    return 'Satiated.'

class Pipeline(object):

  def __init__(self):
    self.ingestion = IngestionStage()
    self.digestion = DigestionStage()

  def run(self):
    ingestion_output = self.ingestion.run()
    digestion_output = self.digestion.run()
    return ingestion_output + digestion_output

if __name__ == '__main__':
  fire.Fire(Pipeline)
```

and accordingly the output should be

```bash
$ python example.py run
Ingesting! Nom nom nom...Burp!
```

I would like to know which change @dbieber would like.
Thank you.

dbieber:
Thanks for noticing the discrepancy.
How about if we make Pipeline.run return a list [ingestion_output, digestion_output].
Then the stage outputs will be displayed on consecutive lines as in the guide's text.

wchliao:
Sounds great!
I'll make a change according to your suggestion.

Janeenm:
,


Sent from Yahoo Mail for iPhone


On Friday, April 23, 2021, 3:07 PM, Wei-Chung Liao ***@***.***> wrote:




Sounds great!
I'll make a change according to your suggestion.

PR #334, opened 2021-04-23
Modify the grouping example to fit the output

wchliao:
I make a change according to #333.
Tested it and everything is correct.

@dbieber could you please review it?
Thank you.


dbieber (approved):

dbieber:
Merging. The advantage of this over what we did in #387 is this conveys consistently that the result is printed and returning data is a reliable way to simultaneously build a python library and a CLI.

Issue #188, opened 2019-08-21
Option to print help to std out


cbrochtrup:
Thanks for creating Fire. It's a fantastic argument parsing tool.

A minor request: I prefer my help strings to be printed to stdout instead of a console like `less`. When the output is sent to a `less`-like program you can't reference the script argument information when trying to type your next command.

The naive way to allow this option is adding a switch so the user can choose to send information console package or to stdout (code below).

https://github.com/google/python-fire/blob/d77453938a4b7a5a2bb71d9cb40397ee8bbc2e0a/fire/core.py#L171
to something like
```
if print_std:
   print(text, file=out)
else:
   console_io.More(
```

I don't see an elegant way to set `print_std` in your source code since you are not accepting contributions to your console package and the `Display` function is not a method of the Fire class. I don't see a straightforward way to have the user set this display option. Obviously, you know the code better than me so you may know a good way to add this option.

I am using fire version 0.2.1 and have seen this behavior with Ubuntu 16.04 and 18.04.

Thank you for your time!

dbieber:
Thanks for this feedback. Let me think on this some and get back to you.

nikking:
One example to look at is `git`. The default for `--help` is stdout, but for commands that are expected to be longer than a page (e.g. `git --diff`), the pager is used.

Git also allows you to adjust the behavior with options (`--no-pager`), an environment variable (`GIT_PAGER`), and configuration (`core.pager`).

Sources: https://www.git-scm.com/docs/git/1.7.4 and https://www.git-scm.com/docs/git-config

jaredtrog:
Any progress on this enhancement? I'd be happy to do the work if @dbieber or other members have some guidance on how they'd like to see this feature implemented.

It looks like @cbrochtrup may not be too far off on the implementation since `More()` currently calls `out.write` if the session is not interactive. Perhaps a function in `console_io` called `StdOut` that more or less mirrors `More`? It could handle checking terminal capabilities to ensure ANSI escape sequences are supported or bail in a more graceful way.

At that point the environment functionality @nikking mentioned could be implemented in core.

nlarusstone:
This is a feature I would really appreciate! I have been enjoying the simplicity of Fire so far, but when you have even a couple steps of nesting it quickly gets frustrating to keep bringing up a pager and having to remember all the arguments. I would love to have my history printed to std out so that I can refer back to it.

Let me know if I can help at all!

dbieber:
[...]

PR #345, opened 2021-07-07
Add custom formatter for Fire result

### LINKED ISSUE #344: Override Result Formatting for pretty and non-intrusive CLIs 🌈✨
Fire's main benefit is that it can be dropped in and turn anything into a CLI! But sometimes there are certain types that either don't play nice with fire's formatters and some that just don't print out in a pretty way.

**What I'm proposing:** Allow users to provide a function that can override result formatting. Simply, it takes the output of the fire call and passes it through the user formatter, and then the result that call is treated as the result of the original fire call. 

### Outputting tabular data:
```python
import random
def data():
    return [
        {
            'is_up': random.choice([False, True]), 
            'value_A': random.random(), 
            'value_B': random.random() + 100,
            'id': random.randint(1000, 5000)} 
        for i in range(8)
    ]

import fire
fire.Fire(data)
```
Outputs this:
```
{"is_up": true, "value_A": 0.6004291859538904, "value_B": 100.77910907893889, "id": 474}
{"is_up": false, "value_A": 0.1406617230697117, "value_B": 100.35721554966845, "id": 740}
{"is_up": true, "value_A": 0.3612392830626744, "value_B": 100.60814663568802, "id": 509}
{"is_up": false, "value_A": 0.11247653550250092, "value_B": 100.2673181440675, "id": 305}
{"is_up": false, "value_A": 0.9505598630828166, "value_B": 100.84615141986525, "id": 85}
{"is_up": true, "value_A": 0.17544933002396768, "value_B": 100.66062056951291, "id": 385}
{"is_up": false, "value_A": 0.25245927587860695, "value_B": 100.75492369068093, "id": 923}
{"is_up": true, "value_A": 0.9237200249249168, "value_B": 100.94228120845642, "id": 702}
```
But if we can define a formatter.
```python
import tabulate

def fancy_table(result):
    if not result:  # show a message instead of an empty response
        return 'nada. sorry.'
    
    # display a list of dicts as a table
    if isinstance(result, (list, tuple)) and all(isinstance(x, dict) for x in result):
        return tabulate.tabulate([
            {col: cell_format(value) for col, value in row.items()}
            for row in result
        ], headers="keys")

    return result  # otherwise, let fire handle it

def cell_format(value, decimals=3, bool=('🌹', '🥀')):
    if value is True:
        return bool[0]
    if value is False:
        return bool[1]
    if isinstance(value, float):
        return '{:.{}f}'.format(value, decimals)
    return value

fire.Fire(data, formatter=fancy_table)
```
Outputs this:
```
is_up      value_A    value_B    id
-------  ---------  ---------  ----
🌹           0.115    100.013  1821
🌹           0.439    100.167  4242
🥀           0.68     100.345  2937
🥀           0.074    100.119  4675
🌹           0.189    100.462  4571
🌹           0.221    100.342  1522
🌹           0.02     100.452  2363
🥀           0.023    100.812  2433
```

**And to address the potential comment that you can do this already by wrapping the function:** yes that's true for simple cases, but if you start wrapping classes or a family of functions (`fire.Fire()` can call any defined function) it becomes less realistic and then `fire` needs wrappers for all of those functions which makes it much more cumbersome (I'm literally doing this in a project and it's such a pain). And that doesn't get into how much more difficult it'd be to wrap object methods that return other objects with their own methods, etc. And on top of that, wrappers mess with function signature inspection which is used by fire for command line help strings so it's problematic in many ways.

This is a very simple addition to fire that can supercharge the look of your CLI with very little effort and intervention in existing code!

Real Use Cases that I've personally encountered:
 - Rest API class (A class that wraps a collection of API methods): provide API result data in pretty ways (e.g. like the table above!!). Right now I'm doing it manually by subclassing it and wrapping every single method (UGH!)
 - Keras models: they do not print out nicely when returned from Fire and I have a file full of functions that return them. It would be a big hassle to wrap them all when I just want to test a couple out quickly

beasteers:
Fixes #344 (see issue for more details)

This lets you define a function that will take the result from the Fire component and allows the user to alter it before fire looks at it to render it. 

**Why?** Because you want to define a global way of formatting your data for your CLI across a family of functions/classes where it is impractical (and a **major pain**) to wrap them all.

### Outputting tabular data:
```python
import random
def data():
    return [
        {
            'is_up': random.choice([False, True]), 
            'value_A': random.random(), 
            'value_B': random.random() + 100,
            'id': random.randint(1000, 5000)} 
        for i in range(8)
    ]

import fire
fire.Fire(data)
```
Outputs this:
```
{"is_up": true, "value_A": 0.6004291859538904, "value_B": 100.77910907893889, "id": 474}
{"is_up": false, "value_A": 0.1406617230697117, "value_B": 100.35721554966845, "id": 740}
{"is_up": true, "value_A": 0.3612392830626744, "value_B": 100.60814663568802, "id": 509}
{"is_up": false, "value_A": 0.11247653550250092, "value_B": 100.2673181440675, "id": 305}
{"is_up": false, "value_A": 0.9505598630828166, "value_B": 100.84615141986525, "id": 85}
{"is_up": true, "value_A": 0.17544933002396768, "value_B": 100.66062056951291, "id": 385}
{"is_up": false, "value_A": 0.25245927587860695, "value_B": 100.75492369068093, "id": 923}
{"is_up": true, "value_A": 0.9237200249249168, "value_B": 100.94228120845642, "id": 702}
```
But if we can define a formatter.
```python
import tabulate

def fancy_table(result):
    if not result:  # show a message instead of an empty response
        return 'nada. sorry.'
    
    # display a list of dicts as a table
    if isinstance(result, (list, tuple)) and all(isinstance(x, dict) for x in result):
        return tabulate.tabulate([
            {col: cell_format(value) for col, value in row.items()}
            for row in result
        ], headers="keys")

    return result  # otherwise, let fire handle it

def cell_format(value, decimals=3, bool=('🌹', '🥀')):
    if value is True:
        return bool[0]
    if value is False:
        return bool[1]
    if isinstance(value, float):
        return '{:.{}f}'.format(value, decimals)
    return value

fire.Fire(data, formatter=fancy_table)
```
Outputs this:
```
is_up      value_A    value_B    id
-------  ---------  ---------  ----
🌹           0.115    100.013  1821
🌹           0.439    100.167  4242
🥀           0.68     100.345  2937
🥀           0.074    100.119  4675
🌹           0.189    100.462  4571
🌹           0.221    100.342  1522
🌹           0.02     100.452  2363
🥀           0.023    100.812  2433
```

Using a formatter means that:
 - if you don't provide a formatter, or if your formatter returns the original value (`lambda x: x`) then nothing is changed
 - if you want to change how some value is displayed, you can just return what you want it to render as
   e.g. `lambda x: {'idk why, but': x} if isinstance(x, list) else x`
 - if you are unable to make a class `__str__` representation look like you want it to, you can handle it in the formatter instead
   e.g. `lambda x: custom_str(x) if isinstance(x, some_class) else x`
 - if you want to display nested dictionaries as yaml, you can use `yaml.dump` as a formatter
 - you can suppress output  e.g. `lambda x: None`
 - you can handle printing yourself by printing inside the formatter and returning None to suppress fire formatting

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -249,7 +249,9 @@ def _PrintResult(component_trace, verbose=False, serialize=None):
 
   # Allow users to modify the return value of the component and provide 
   # custom formatting.
-  if callable(serialize):
+  if serialize:
+    if not callable(serialize):
+      raise FireError("serialize must be callable.")
     result = serialize(result)
 
   if value_types.HasCustomStr(result):
--- a/fire/core.py
+++ b/fire/core.py
@@ -78,7 +78,7 @@ if six.PY34:
   import asyncio  # pylint: disable=import-error,g-import-not-at-top  # pytype: disable=import-error
 
 
-def Fire(component=None, command=None, name=None, formatter=None):
+def Fire(component=None, command=None, name=None, serialize=None):
   """This function, Fire, is the main entrypoint for Python Fire.
 
   Executes a command either from the `command` argument or from sys.argv by
@@ -164,7 +164,7 @@ def Fire(component=None, command=None, name=None, formatter=None):
     raise FireExit(0, component_trace)
 
   # The command succeeded normally; print the result.
-  _PrintResult(component_trace, verbose=component_trace.verbose, formatter=formatter)
+  _PrintResult(component_trace, verbose=component_trace.verbose, serialize=serialize)
   result = component_trace.GetResult()
   return result
 
@@ -241,7 +241,7 @@ def _IsHelpShortcut(component_trace, remaining_args):
   return show_help
 
 
-def _PrintResult(component_trace, verbose=False, formatter=None):
+def _PrintResult(component_trace, verbose=False, serialize=None):
   """Prints the result of the Fire call to stdout in a human readable way."""
   # TODO(dbieber): Design human readable deserializable serialization method
   # and move serialization to its own module.
@@ -249,8 +249,8 @@ def _PrintResult(component_trace, verbose=False, formatter=None):
 
   # Allow users to modify the return value of the component and provide 
   # custom formatting.
-  if callable(formatter):
-    result = formatter(result)
+  if callable(serialize):
+    result = serialize(result)
 
   if value_types.HasCustomStr(result):
     # If the object has a custom __str__ method, rather than one inherited from
--- a/fire/core.py
+++ b/fire/core.py
@@ -78,7 +78,7 @@ if six.PY34:
   import asyncio  # pylint: disable=import-error,g-import-not-at-top  # pytype: disable=import-error
 
 
-def Fire(component=None, command=None, name=None):
+def Fire(component=None, command=None, name=None, formatter=None):
   """This function, Fire, is the main entrypoint for Python Fire.
 
   Executes a command either from the `command` argument or from sys.argv by
@@ -164,7 +164,7 @@ def Fire(component=None, command=None, name=None):
     raise FireExit(0, component_trace)
 
   # The command succeeded normally; print the result.
-  _PrintResult(component_trace, verbose=component_trace.verbose)
+  _PrintResult(component_trace, verbose=component_trace.verbose, formatter=formatter)
   result = component_trace.GetResult()
   return result
 
@@ -241,11 +241,16 @@ def _IsHelpShortcut(component_trace, remaining_args):
   return show_help
 
 
-def _PrintResult(component_trace, verbose=False):
+def _PrintResult(component_trace, verbose=False, formatter=None):
   """Prints the result of the Fire call to stdout in a human readable way."""
   # TODO(dbieber): Design human readable deserializable serialization method
   # and move serialization to its own module.
   result = component_trace.GetResult()
+
+  # Allow users to modify the return value of the component and provide 
+  # custom formatting.
+  if callable(formatter):
+    result = formatter(result)
 
   if value_types.HasCustomStr(result):
     # If the object has a custom __str__ method, rather than one inherited from
--- a/fire/core.py
+++ b/fire/core.py
@@ -251,7 +251,7 @@ def _PrintResult(component_trace, verbose=False, serialize=None):
   # custom formatting.
   if serialize:
     if not callable(serialize):
-      raise FireError("serialize must be callable.")
+      raise FireError("serialize argument {} must be empty or callable.".format(serialize))
     result = serialize(result)
 
   if value_types.HasCustomStr(result):
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -200,14 +200,20 @@ class CoreTest(testutils.BaseTestCase):
         return ', '.join(str(xi) for xi in x)
       if isinstance(x, dict):
         return ', '.join('{}={!r}'.format(k, v) for k, v in x.items())
+      if x == 'special':
+        return 'SURPRISE!!'
       return x
+
+    ident = lambda x: x
     
     with self.assertOutputMatches(stdout='a, b', stderr=None):
-      result = core.Fire(lambda x: list(x), command=['[a,b]'], formatter=formatter)
+      result = core.Fire(ident, command=['[a,b]'], formatter=formatter)
     with self.assertOutputMatches(stdout='a=5, b=6', stderr=None):
-      result = core.Fire(lambda x: dict(x), command=['{a:5,b:6}'], formatter=formatter)
+      result = core.Fire(ident, command=['{a:5,b:6}'], formatter=formatter)
     with self.assertOutputMatches(stdout='asdf', stderr=None):
-      result = core.Fire(lambda x: str(x), command=['asdf'], formatter=formatter)
+      result = core.Fire(ident, command=['asdf'], formatter=formatter)
+    with self.assertOutputMatches(stdout='SURPRISE!!', stderr=None):
+      result = core.Fire(ident, command=['special'], formatter=formatter)
 
 
   @testutils.skipIf(six.PY2, 'lru_cache is Python 3 only.')
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -194,6 +194,22 @@ class CoreTest(testutils.BaseTestCase):
         7,
     )
 
+  def testCustomFormatter(self):
+    def formatter(x):
+      if isinstance(x, list):
+        return ', '.join(str(xi) for xi in x)
+      if isinstance(x, dict):
+        return ', '.join('{}={!r}'.format(k, v) for k, v in x.items())
+      return x
+    
+    with self.assertOutputMatches(stdout='a, b', stderr=None):
+      result = core.Fire(lambda x: list(x), command=['[a,b]'], formatter=formatter)
+    with self.assertOutputMatches(stdout='a=5, b=6', stderr=None):
+      result = core.Fire(lambda x: dict(x), command=['{a:5,b:6}'], formatter=formatter)
+    with self.assertOutputMatches(stdout='asdf', stderr=None):
+      result = core.Fire(lambda x: str(x), command=['asdf'], formatter=formatter)
+
+
   @testutils.skipIf(six.PY2, 'lru_cache is Python 3 only.')
   def testLruCacheDecoratorBoundArg(self):
     self.assertEqual(
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -194,8 +194,8 @@ class CoreTest(testutils.BaseTestCase):
         7,
     )
 
-  def testCustomFormatter(self):
-    def formatter(x):
+  def testCustomSerialize(self):
+    def serialize(x):
       if isinstance(x, list):
         return ', '.join(str(xi) for xi in x)
       if isinstance(x, dict):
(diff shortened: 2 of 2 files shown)


beasteers:
@dbieber I was wondering if I could get your opinion on this because I've run into this on numerous occasions and I think this would be a great addition to `fire`! it might even provide room for developing `fire` themes eventually!

luxiaba:
can't agree more, as mentioned [here](https://github.com/google/python-fire/blob/ed44d8b801fc24e40729abef11b2dcbf6588d361/fire/core.py#L381-L383), after each iteration of parsing, the component is updated to another new object, so there is no way to uniformly process the final result (maybe we need to remove some keys in the final dict or beautify the final result).

actually,  i think it might be better to let users to provide a customized print method instead of a formatter. At present, `_PrintResult` mainly uses `print' without any parameters, and there is no way to interfere with the behavior of print (such as remove new line: end='').

beasteers:
Glad other people find this useful too! 

That is an interesting idea. I still like the idea of a formatter because it allows you to do a minimal override and pre-process the message if you want to, but still rely on the built in output mechanisms for everything else - so if you override it for one type, builtin formatting for everything else will still work as normal.

```python
def my_formatter(x):
    # handle a class that doesn't have a nice string output
    if isinstance(x, tf.keras.Model):  # like a keras model
        x.summary()  # prints a nice output
        return  # return empty so the default object rendering doesn't happen
    return x  # everything else is handled normally, e.g. dicts, lists, objects
```

But what you're talking about is possible with the formatter argument in this PR. You can just print from inside the formatter function and then return `None` (or don't return anything) and it'll work just like your `_PrintResult` suggestion.

```python
def my_formatter(x):
    print(x, end=' ')  # print instead of return and you can add print arguments
```

luxiaba:
aha, here `my_formatter` is like a customized print method(at least it seems did the job of printing, instead of always returning something or influencing others obj like a formatter, as for `_PrintResult `, maybe we can use it as a default printer):

*[before](https://github.com/google/python-fire/blob/ed44d8b801fc24e40729abef11b2dcbf6588d361/fire/core.py#L166-L169):*
```python
# The command succeeded normally; print the result.
_PrintResult(component_trace, verbose=component_trace.verbose)
result = component_trace.GetResult()
return result
```

*after:*
```python
# The command succeeded normally; print the result.
result = component_trace.GetResult()

# show result
if has_customized_printer():
  customized_printer(result, component_trace.verbose)
else:
  # here we can separate the helper display from `_PrintResult`, let it keep the simple display function.
  _PrintResultWithoutHepler(result, verbose=component_trace.verbose)

# show helper
if should_show_help():
  show_help(component_trace, component_trace.verbose)
return result
```

Anyway, it's just a little bit of my personal thought, and I think your unifed formatter is already great idea, hope to merged ASAP☺️

beasteers:
Hmm not sure if I understand what your suggestion is 🤔

Ya the custom `formatter` is meant to be really flexible and can already handle both of our use-cases. (You don't have to change anything in this PR for your `after:` code to work the way you intended. By returning nothing from the formatter, you will be bypassing fire's outputting, just like you wanted.

Yeah I hope it gets merged soon too! Thanks for your input!

dbieber:
Yes, this is a welcome change!
Note unfortunately the time I have to review and merge is limited now; it'll likely not be until early next year that we get this checked in.

Have a look at https://github.com/google/python-fire/issues/188#issuecomment-596780455
for an older discussion about how we want to do this.

Summary of top-level thoughts:
* Let's call the argument `display` rather than formatter, for consistency with other arguments we might add in the future (e.g. serialize)
* I like the approach of allowing the display function to "pass through" to the default by returning a value.
  * However, if the display function doesn't return a value, we still want `result` to be set correctly in interactive mode.
* Doesn't need to be in the first pass, but I want to support multiple possible signatures for the argument to display so that the five use-cases at the start of [the comment](https://github.com/google/python-fire/issues/188#issuecomment-596780455) all work.

beasteers:
> Note unfortunately the time I have to review and merge is limited now; it'll likely not be until early next year that we get this checked in.

yep no worries! I probably wouldn't be able to work on it until then anyways either.

> Let's call the argument display rather than formatter, for consistency with other arguments we might add in the future (e.g. serialize)

love it, sounds much better ✨

> However, if the display function doesn't return a value, we still want result to be set correctly in interactive mode.

I've never used interactive mode so I'll need to dig in to see what this means, but the display function doesn't replace the result value used in chaining. So regardless of what the display function returns, the fire chaining will still work the same way because it still uses the original value. (the format call happens inside the _PrintResult call which has no return value).

> Doesn't need to be in the first pass, but I want to support multiple possible signatures for the argument to display so that the five use-cases at the start of the comment all work.

Hmm maybe we could have `format=fancy_table` and `display=print`? It may be easier to decouple the output mode (less/print) with the result processing so you can mix and match.

beasteers:
Wait actually reading thru that other thread more, this PR is actually providing `serialize` as you are describing it there (not `display`), because I designed it with this use-case in mind:

```python
def json_format(x):
    return json.dumps(x, indent=4)

fire.Fire(..., format=json_format)
```
as well as csv, yaml, (fancy_table is also just a serialization) etc...

Like this was never meant as a display function that needs to manage printing out anything (though you are free to do it as a side effect if you want)

It's primarily just to provide a method for custom serialization, with the ability to fallback to default serialization implicitly (by returning something other than None/str).

Also with this method, if you want to combine multiple formatters that work for different types, all you'd have to do is:
```python
fire.Fire(..., format=lambda x: (
        list_of_dicts_as_table(
            list_of_lists_as_csv(
                dicts_as_yaml(x)))
))
```
which should work out of the box currently. But this composition would not be possible without the pass-through mechanism.

dbieber:
Yes! That's exactly the direction I was thinking in:

The user can specify either (or both) of two arguments (as you saw, I'm thinking to call them `serialize` and `display`):

The rough signature of serialize is `serialize(result) -> output` (no side-effects).
The rough signature of display is `display(output) -> None` (with the side-effect of displaying the output).

For your use-case, you could in principle use either serialize or display, but like you say I think serialize is more appropriate. If you use serialize, the recommended approach would be to return the table as a string. If you use display, then you would have to print the string yourself rather than returning it.

beasteers:
Ok cool! In that case I don't think there's much more to do other than rename formatter= to serialize= and maybe add a couple builtin serializers. 

Idk if we should do display= as a separate PR? I'm less familiar with the nuances around output and paging that ppl are requesting.

(Obvs we can decide that down the road when we both have time)

dbieber:
Yes, adding them in separate PRs would be fine.

beasteers:
sounds good! yeah I'll probably let someone more motivated by the display overriding handle it since I don't have much of an opinion about how it should be dealt with.

Cool this is exciting tho!

fire/core.py
```diff
@@ -241,12 +241,17 @@ def _IsHelpShortcut(component_trace, remaining_args):
...
+def _PrintResult(component_trace, verbose=False, serialize=None):
   """Prints the result of the Fire call to stdout in a human readable way."""
   # TODO(dbieber): Design human readable deserializable serialization method
   # and move serialization to its own module.
   result = component_trace.GetResult()
 
+  # Allow users to modify the return value of the component and provide 
+  # custom formatting.
+  if callable(serialize):
+    result = serialize(result)
```
dbieber, line 253: We'll want to handle the not-callable case too.
I think if we cannot call serialize, we raise a FireError.

fire/core_test.py
```diff
@@ -194,6 +194,28 @@ def testClassMethod(self):
...
     )
 
+  def testCustomSerialize(self):
+    def serialize(x):
+      if isinstance(x, list):
+        return ', '.join(str(xi) for xi in x)
+      if isinstance(x, dict):
+        return ', '.join('{}={!r}'.format(k, v) for k, v in x.items())
+      if x == 'special':
+        return ['SURPRISE!!', "I'm a list!"]
```
dbieber, line 204: The name "serialize" implies that the output of serialize is either bytes or a string.
But that's not what we're requiring.
I wonder if there's a better name (maybe that's "format" after all.)
I'll think on this, but curious for your thoughts.

fire/core.py
dbieber, line 253: We'll want to leave result unchanged, and introduce a new variable serialized_result that we use from here on out.
This way in interactive mode (-- -i) the user will still have access to the non-serialized result, as well as serialized_result.

fire/core_test.py
dbieber: Maybe "process" or "finalize"...
I'll keep thinking.

dbieber: Forgive my unpolished thinking aloud:

Maybe "serialize" is fine, and the logic is: "if you don't finish serializing it, we'll continue serializing it for you using the default serializer."
And then "display" will only ever be given bytes or a string, rather than being directly given the output of serialize.
One challenge with this is that it disallows different serialization for different display modes.
We could (in a subsequent PR) allow serialize to accept the "kind" parameter too to reenable this capability.

fire/core.py
beasteers: Maybe we just do:

```python
if serialize:
    result = serialize(result)
```
which will just raise whatever error the object would raise if tried to be called (probably typeerror)

fire/core_test.py
beasteers: side note: one thing that would be nice is if we could have some sort of override via the CLI where the user could do `--format json` or `--format csv` or something like that

fire/core.py
beasteers: This is internally scoped inside `_PrintResult` and doesn't actually affect the value of the interactive result, unless I'm missing some weird introspection thing that's happening

dbieber: My bad. You're right, it's fine as is.

(The relevant place is `variables['result'] = component` at line 611, which as you say isn't affected by what's happening here.)

fire/core_test.py
beasteers: Some notes about the naming - I like `serialize` and `format` because they both imply a string result. I think there might be a problem with `process` or `finalize` cuz they could imply that you have the ability to modify the result after each iteration (after each getattr / getitem / call).

I can't think of any other words better than those two tho.

> Maybe "serialize" is fine, and the logic is: "if you don't finish serializing it, we'll continue serializing it for you using the default serializer."

I think this is fine because even if the user doesn't realize that they can't return a string from it, it's not going to break anything, it might just make a little more work for them.

fire/core.py
beasteers: addressed - it will not raise `FireError("serialize argument {} must be empty or callable.".format(serialize))`

dbieber:
This looks good! I'm going to squash-and-merge it in now.

The next steps are:
* Confirm interactive mode works as expected even when serialize= is set.
* Add a default serializer function as in #188 
* Add the display= argument next

Thanks again for the PR!

dbieber:
The release went out today. This is the most significant change in the release and is highlighted in the [release notes](https://github.com/google/python-fire/releases/tag/v0.5.0).
Thanks for your patience, and of course thanks for the contribution as well!

Issue #327, opened 2021-03-19
Parse arguments based on type hints in the corresponding function signature


dhimmel:
`add_argument` from `argparse` has a `type` [option](https://docs.python.org/3/library/argparse.html#type) that "allows any necessary type-checking and type conversions to be performed". The ability to ensure the CLI args are converted to certain types helps ensure consistent CLI behavior and avoid bugs.

Since fire strives to be as automated as possible, it seems natural that it could inspect the type hints for the function signature that is being fired and cast the arguments to these types. If the arguments can't be converted to the type signature, an error would be raised to alert the CLI user to the invalid argument.

The [docs are clear](https://google.github.io/python-fire/guide/#argument-parsing) that this is not the current design:

> The types of the arguments are determined by their values, rather than by the function signature where they're used.

But the convenience of relying on type hints for argument type conversion would be major. It could be implemented in an opt-in manner for back compat. Would this idea ever be considered? Are there any CLI projects that are already doing this? 

See also the related issue at https://github.com/google/python-fire/issues/260.

zqqqqz2000:
You can try [afire](https://github.com/zqqqqz2000/afire), which is a fork of python-fire supporing type convert according to the type hint.

PR #350, opened 2021-08-19
Add decorator to parse function type hints

keanpantraw:
This allows to add ability to tell fire what types are expected by type hints (see #327 ) and not change current behaviour by making it optional decorator.


keanpantraw:
@googlebot I fixed it.

Issue #225, opened 2020-02-19
Get a list of all valid commands when writing --help


bfelbo:
Thanks for creating this amazing library. I'm somewhat confused as to how the `--help` command is implemented. When running `python example.py --help` Fire only prints the flags, not the actual commands that can be run (see example below):

```
NAME
    fury.py

SYNOPSIS
    fury.py <flags>

FLAGS
    --verbosity=VERBOSITY
```

How do I make `--help` show all the possible commands and why is that not the default? :)

dbieber:
Hi @bfelbo,

Are you suggesting that the help show:

```
fury.py --verbosity=VERBOSITY
```

or is there a command that's actually _missing_ from the help, that you'd like it to show?

It should be showing all the possible commands, but not listed out as all possible complete commands explicitly simply due to formatting considerations. In particular, all commands share a prefix, and sometimes that prefix is quite long, so we choose not to write out that prefix repeatedly. Always open to suggestions for improvement though.

bfelbo:
Thanks for the response. I'll make it more clear :) See the below example from the docs.
```
import fire

class Calculator(object):

  def add(self, x, y):
    return x + y

  def multiply(self, x, y):
    return x * y

if __name__ == '__main__':
  fire.Fire(Calculator)
```
Right now running `python example.py --help` returns the following:
```
NAME
    example.py

SYNOPSIS
    example.py
```
I think it'd be great if running help would actually tell you what CLI commands are available, i.e. `add` and `multiply` in this case. Otherwise, how can you figure out what you can do with the CLI? Maybe I'm missing something obvious here :)

bfelbo:
> It should be showing all the possible commands

This doesn't seem to be the case? (see my previous comment and example)

buzzlaker:
@bfelbo As far as I know, it doesn't generate the options for you. You're saying it would be nice if help also showed something like this?
```
add - examply.py add x y
multiply - examply.py multiply x y
```

I don't think that would be too hard to add-in.

bfelbo:
Exactly. That would be amazing!

For the class-based interface, I would probably ignore the `self` argument. It would then become something like the following:

```
NAME
    example.py

COMMANDS
    add x y
    multiply x y

SYNOPSIS
    example.py
```

bfelbo:
If the arguments have a default, it'd be great to include them too, e.g.:

```
NAME
    example.py

COMMANDS
    add x y
    multiply x y
    root x z=2

SYNOPSIS
[...]

PR #364, opened 2021-12-06
Fix Get a list of all valid commands when writing --help

Rohan-Salwan:
- fixed: #225

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -326,7 +326,7 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
       # Methods and properties should only be accessible on instantiated
       # objects, not on uninstantiated classes.
       if class_attr.kind in ('method', 'property'):
-        return False
+        return True
       # Backward compatibility notes: Before Python 3.8, namedtuple attributes
       # were properties. In Python 3.8, they have type tuplegetter.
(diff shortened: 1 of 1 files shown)


fire/completion.py
```diff
@@ -326,7 +326,7 @@ def MemberVisible(component, name, member, class_attrs=None, verbose=False):
       # Methods and properties should only be accessible on instantiated
       # objects, not on uninstantiated classes.
       if class_attr.kind in ('method', 'property'):
-        return False
+        return True
```
dbieber, line 329: Can you provide an example of why you want to make this change?
The comment at 326-327 suggests that False was intended.

If we are changing this to True, we'll want to change that comment as well.

dbieber: Sorry, I see you already have done so in #225. My bad.

dbieber:
Note to self: This makes the change discussed at https://github.com/google/python-fire/issues/225#issuecomment-590512721

bwtaylor (commented):
The comments on line 326-327 should be modified to say that we do want methods and properties of classes to be accessible so that --help to see them.

bwtaylor (commented):
This will show class attributes that start with _, an inconsistency with how it works for objects. This can be easily fixed via:
  return not name.startswith('_')

Issue #329, opened 2021-03-30
Less than helpful help text for Optional[int] argument


genos:
Hi friends! Thanks as always for this _amazing_ library; it's really made building CLIs in Python a breeze.

One thing I've noticed recently: when `Fire`-ing a function with an argument of type `Optional[int]`, the help text for that argument has a bit of extraneous noise and gets cut off in such a way that we never see the `int` portion. Exempli gratia:

### Environment

```
~/tmp ∃ python3 -m venv env
~/tmp ∃ source env/bin/activate
(env) ~/tmp ∃ pip install -U -q pip fire
(env) ~/tmp ∃ python --version
Python 3.8.7
(env) ~/tmp ∃ pip freeze
fire==0.4.0
six==1.15.0
termcolor==1.1.0
```

### Script
```python
#!/usr/bin/env python
from typing import Optional
from fire import Fire

def main(n: Optional[int] = None):
    n = 5 if n is None else n
    print(f"You chose {n}.")

if __name__ == "__main__":
    Fire(main)
```

### Running

When I run `./test.py --help`, I receive the following help message:

```
NAME
    test.py

SYNOPSIS
    test.py <flags>

FLAGS
    --n=N
        Type: Optional[typing.Unio...
        Default: None
```

### Desiderata

Perhaps it's happening because [`Optional[T]` is equivalent to `Union[T, None]`](https://docs.python.org/3/library/typing.html#typing.Optional), but I find the type annotation's abbreviation`Optional[typing.Unio...]` in the help text a little wanting; ideally I'd prefer something like `Optional[int]` there. Granted, if this weren't a contrived, simplified example, I'd include a docstring that described the optional integer argument and not expect `fire` to take care of everything for me! But it'd be nice if we could see the type parameter inside that `Optional`.

Thanks again!

damienjburks:
I'd like to do this one.

PR #380, opened 2022-01-16
adding code to support optional type appreviation request

damienjburks:
Addresses #329


damienjburks:
Could you please submit a rescan for the CLA? Thanks!
image

fire/inspectutils.py
```diff
@@ -143,7 +144,7 @@ def Py3GetFullArgSpec(fn):
 
   if sig.return_annotation is not sig.empty:
     annotations['return'] = sig.return_annotation
-
+
```
genos, line 147: Extra whitespace

fire/inspectutils.py
```diff
@@ -165,6 +166,10 @@ def Py3GetFullArgSpec(fn):
       varkw = name
     if param.annotation is not param.empty:
       annotations[name] = param.annotation
+      if "Union" in str(annotations[name]):
```
genos, line 169: This works for the example from #329 in Python 3.8, but

1. I think this will unfortunately not work for other types of `Union`s (e.g. `Union[str, int]` instead of `Option[int] == Union[int, None]`); and
2. it doesn't work in Python 3.9:

```
~/tmp ∃ python3 -m venv env
~/tmp ∃ source env/bin/activate
(env) ~/tmp ∃ pip install -U -q pip fire
(env) ~/tmp ∃ python --version
Python 3.9.9
(env) ~/tmp ∃ pip freeze
fire==0.4.0
six==1.16.0
termcolor==1.1.0
(env) ~/tmp ∃ cat test.py
#!/usr/bin/env python
from typing import Optional
from fire import Fire

def main(n: Optional[int] = None):
    n = 5 if n is None else n
    print(f"You chose {n}.")

if __name__ == "__main__":
    Fire(main)%
(env) ~/tmp ∃ ./test.py --help
INFO: Showing help with the command 'test.py -- --help'.

NAME
    test.py

SYNOPSIS
    test.py <flags>

FLAGS
    --n=N
        Type: Optional[typing.Optional[int]]
        Default: None
```

fire/inspectutils.py
```diff
@@ -165,6 +165,14 @@ def Py3GetFullArgSpec(fn):
       varkw = name
     if param.annotation is not param.empty:
       annotations[name] = param.annotation
+      if "Optional" in str(annotations[name]) or "Union" in str(annotations[name]):
+        tuple_param1 = get_args(annotations[name])[0].__name__
+        tuple_param2 = get_args(annotations[name])[1].__name__
+        
+        if str(tuple_param2) != "NoneType":
+          annotations[name] = tuple_param1 + ", " + tuple_param2
```
genos, line 173: What if it's a `Union` of more than two types? Maybe something like

```python
annotations[name] = ", ".join(x.__name__ for x in get_args(annotations[name]))
```

damienjburks:
Added that adjustment. I wasn't aware that Union could have more than 2 options, so that you for the recommendation @genos

dbieber:
Thanks for the contribution! This looks good.
What Python versions does this support?

damienjburks:
@dbieber it supports python versions 3.x and up. I have not tested this with any of the python 2 versions.

Let me know if this is an issue.

dbieber:
We don't need the improvement to apply to Python 2, but we do need to make sure we don't break fire for Python 2.7.

damienjburks:
@dbieber I will check into this, and attach my results to this issue. Thanks.

koconder (approved):

PR #391, opened 2022-04-29
Update core.py

ordas21:
This should make the logging for robust and allow for the suppressing and redirection of messages

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -61,6 +61,7 @@ import re
 import shlex
 import sys
 import types
+import logging
 
 from fire import completion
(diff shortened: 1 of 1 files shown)


koconder (approved):

dbieber:
Thanks for the PR.

I'm not familiar with the logging module enough to know what benefit this provides, or what behavior changes it entails. I have not had a chance to look into this yet.

dbieber (commented):
I expect that if we make this change we probably also would want to make some corresponding changes for printing errors at line 302.

PR #395, opened 2022-06-05
chore: light refactoring

koconder:
Some light refactoring and clean-up

* Remove excessive if/else
* f-name function use where possible


dbieber:
Thanks for looking into these changes, however some notes preventing merge:

* We cannot use f-strings since we still support Python 2.7
* I don't view replacing multi-line if/else with a ternary if-else as a strict improvement -- there are readability tradeoffs. i.e. for `_GetCurrentCommand` I think it reduces readability.

PR #399, opened 2022-07-03
docs: Fix a few typos

timgates42:
There are small typos in:
- fire/console/console_attr.py
- fire/custom_descriptions.py

Fixes:
- Should read `support` rather than `suport`.
- Should read `pertinent` rather than `pertinant`.
- Should read `environment` rather than `envrionment`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


dbieber:
Thanks for fixing these typos!

The console/ directory is actually a third_party dep and we don't generally update that code manually (since the updates just get clobbered when we pull in from upstream.) I'll merge your PR as is though (and if/when we next pull in a console update, the typos might return!)

PR #406, opened 2022-09-21
Fix #338 arg parsing truncates at hash

### LINKED ISSUE #338: arg parsing truncates at hash
Fire's arg parser truncates strings with hashes (#) in them.   Simple test:

test.py:
```
import fire
def test(a): print(a, type(a))
fire.Fire(test)
```

```
> python test.py hi#there
hi <class 'str'>
> python test.py "hi#there"
hi <class 'str'>
```

If this is by design, it doesn't appear to be documented anywhere.

xrzqx:
Closes https://github.com/google/python-fire/issues/338

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,0 +1,8 @@
+# import fire
+# import sys
+
+# sys.argv.append('(1,2)')
+# fire.Fire(lambda obj: type(obj).__name__)
+
+import fire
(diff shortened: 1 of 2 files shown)


dbieber:
Thanks for the PR. We'll need tests before we can incorporate this fix. Let me know if this is something you want to do.

PR #424, opened 2022-12-17
Detect the program name when `python -m` was executed

### LINKED ISSUE #76: Use package name when __main__ is used

meowmeowmeowcat:
Closes #76

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -109,7 +109,8 @@ def Fire(component=None, command=None, name=None, serialize=None):
         code 2. When used with the help or trace flags, Fire will raise a
         FireExit with code 0 if successful.
   """
-  name = name or os.path.basename(sys.argv[0])
+
+  name = _GetProgName(name)
 
   # Get args as a list.
   if isinstance(command, six.string_types):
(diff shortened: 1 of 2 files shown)


fire/core.py
```diff
@@ -281,6 +282,45 @@ def _PrintResult(component_trace, verbose=False, serialize=None):
     Display(output, out=sys.stdout)
 
 
+def _GetProgName(name, main=None):
```
dbieber, line 285: In the current PR I don't think the main arg is passed anywhere (other than the test). Is a file missing?

dbieber:
Good idea! Thanks for submitting 👍

fire/core.py
meowmeowmeowcat: It is only passed during testing.

fire/core.py
```diff
@@ -281,6 +282,45 @@ def _PrintResult(component_trace, verbose=False, serialize=None):
...
+
+  This function returns the program name that should be
+  displayed.
+
+  If ``python -m`` was used to execute a module, ``python -m name`` will be
+  returned, instead of ``__main__.py``.
+
+  Args:
+    name: Optional. The name of the command as entered at the command line.
+    main: Optional. This should only be passed during testing.
```
meowmeowmeowcat, line 296: See here :)

dbieber:
The new behavior doesn't look quite like what we want:

```
SYNOPSIS
    '/usr/bin/python -m sample' ARGS
```

Instead, we'd prefer the following, right?
```
SYNOPSIS
    /usr/bin/python -m sample ARGS
```

dukecat0:
It looks like the quotes come from here:
https://github.com/google/python-fire/blob/91febb4df8b4967a5d041110f55884bd95b139c2/fire/trace.py#L190
I think the reason is that there are spaces inside `self.name`, i.e. `/usr/bin/python -m sample`. Any opinions on how should we deal with this case?

PR #428, opened 2023-01-10
adding python 3.10 [tag & CI]

Borda:
I believe that Fire is fully compatible with all new Python versions so:
- adding tags to package
- adding testing for them

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,13 +1,13 @@
 name: Python Fire
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.7, 3.8, 3.9]
+        python-version: ["2.7", "3.5", "3.7", "3.8", "3.9", "3.10"]
 
     steps:
      # Checkout the repo.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Python Fire
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,7 @@ setup(
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
 
         'Operating System :: OS Independent',
         'Operating System :: POSIX',


hugovk (approved):
Looks good.

Might be good to test 3.12-dev as well, to make sure things here are ready, and also to help test CPython itself before the big release in October.

Borda:
> Might be good to test 3.12-dev as well, to make sure things here are ready, and also to help test CPython itself before the big release in October.

sure, just would make it in separate PR to keep this change focus... :chipmunk:

Borda:
Hello @dbieber, could you pls check and approve running CI... :chipmunk:

dbieber:
Thanks for the PR.
Looks like we'll have to adjust the asyncio support for recent Python versions

`E   AttributeError: module 'asyncio' has no attribute 'coroutine'`

Borda:
ok, so lets do it one by one :chipmunk:

dbieber:
What's the significance of the squirrel emoji?

Borda:
> What's the significance of the squirrel emoji?

Nothing special, just mood at the moment, but I can remove it if you wish :otter:

Borda:
Anyways, it seems that the CI is also failing for py3.8, which was untouched in this PR, so maybe something on master? Maybe worse to implement version freeze?

dbieber:
> I can remove it if you wish 🦦

Not at all! Was just curious. 🐙 

> Anyways, it seems that the CI is also failing for py3.8, which was untouched in this PR

Yes, looks like an update to the linter. #430 addresses. But 3.7 is also failing, and that looks like a linter bug, not a fire one!

Borda:
@dbieber, it seems this one #431 make all :green_circle: again

hugovk:
#431 needs a little path fix: please see https://github.com/google/python-fire/pull/433

hugovk:
#433 is merged, ready for rebase/update here.

Borda:
@hugovk would it be possible in the setting to enable:
- Always suggest updating pull request branches 
- Allow auto-merge

hugovk:
It would, but I'm not a member of this project :)

For auto-merge, we'd need to mark some status checks as required. The CLA is an obvious choice. For the CI, it can be tedious having to click through in the admin settings to mark each individual Python version as a required check; and then when adding/dropping versions, having to go back and repeat.

A common solution is to add a final noop job, which depends on the main test matrix job. And then only that final success job needs to be added as a required check. For example: 

* https://github.com/hugovk/norwegianblue/blob/0b5047fdce6339107326473c139b4b5a9741412a/.github/workflows/test.yml#L49-L55
* https://github.com/hugovk/norwegianblue/actions/runs/4061046465
* https://github.com/hugovk/norwegianblue/pull/123

<details>
<summary>Screenshots</summary>


image

image


</details>

Borda:
> A common solution is to add a final noop job, which depends on the main test matrix job. And then only that final success job needs to be added as a required check. For example:
> 
> * [hugovk/norwegianblue@`0b5047f`/.github/workflows/test.yml#L49-L55](https://github.com/hugovk/norwegianblue/blob/0b5047fdce6339107326473c139b4b5a9741412a/.github/workflows/test.yml#L49-L55)
> * [hugovk/norwegianblue/actions/runs/4061046465](https://github.com/hugovk/norwegianblue/actions/runs/4061046465)

yes I did it too but I think your job won't reflect skip/cancel event, see: https://github.com/Lightning-AI/ecosystem-ci/blob/4ed3826392d0b779f8fabb7d559f5d59bb3fd261/.github/workflows/ci_compatible.yml#L112-L124

dbieber:
I merged the version change (CI for 3.10).
I left the trigger as is since I don't understand the security implications of it well enough yet.

PR #431, opened 2023-02-01
freeze CI requirements

Borda:
I would suggest to frezee CI depencies to limit accidental test failing die to unrelated changes... as I see failing tests in https://github.com/google/python-fire/pull/428#issuecomment-1412539465 without touching anything significant
so these versions are set as of Dec 2022 as that was last :green_circle: commit to the main branch
in the next PR #432 we can add dependabot, which is GitHub build-in mechanism to keep dependencies safely bumped...


dbieber:
LGTM

PR #433, opened 2023-02-01
Fix path to requirements.txt

hugovk:
Follow on from https://github.com/google/python-fire/pull/431 to fix the CI.

Sample build: https://github.com/hugovk/python-fire/actions/runs/4068330039

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -19,7 +19,7 @@ set -e
 
 PYTHON_VERSION=${PYTHON_VERSION:-2.7}
 
-pip install -U -r requirements.txt
+pip install -U -r .github/scripts/requirements.txt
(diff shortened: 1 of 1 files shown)


Borda:
wondering how that the PR #431 went :green_circle:

hugovk:
Because it only ran the CLA check:

image

That was because the test workflow only triggered for `push` events, and for a push _from a fork_, means it would only run on the fork's CI, not this upstream:

```yml
on: [push]
```

It failed on https://github.com/google/python-fire/pull/428 because it fixes the triggers:

```yml
on: [push, pull_request]
```

So it also runs on upstream for the PR trigger. This is good, so we can see if PRs from forks pass on the CI. This is very important for open source, where people often contribute from forks.

PR #436, opened 2023-02-09
Codespell fix typos, add workflow for the future to catch new ones

yarikoptic:


dbieber:
Thanks for fixing the typos. Let's not add codespell at this time, but I'd be happy to merge the PR w/ just the typo fixes.

yarikoptic:
You are the boss.  I force pushed kicking away the last commit with the workflow.  Let me know if you would like to get rid of `.codespellrc` as well

dbieber:
> Let me know if you would like to get rid of .codespellrc as well
Yep, thanks

yarikoptic:
IMHO "your loss", but done now.

dbieber:
Thanks Yari! I'm not opposed to adding spell check more generally; just adding codespell would require more research and review on my part and would delay getting things checked in. So, I'm happy to merge the pure-fix PR now rather than having an easy fix get blocked by a slower process.

yarikoptic:
Understood. Let me know if your want sister pr with those

PR #446, opened 2023-03-31
Fix pandas.DataFrame support in core._PrintResult

paul-ada:
+ and add `HasCustomRepr` check in value_types

**Issue**: 
When calling `Fire` on a function that returns a pd.DataFrame, the _PrintResult function in core.py prints the manual for DataFrames from the pandas Docs to stdout. 

Since `if value_types.HasCustomStr(result):` doesn't recognize the DataFrame's custom \_\_str__ as it is simply seemingly inherited from the custom \_\_repr__ for DataFrames (see pandas.core.frame.DataFrame.\_\_repr__), the `_PrintResult` function displays the DataFrame's `helptext.HelpText` instead.

This implementation should fix the issue.

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -262,6 +262,13 @@ def _PrintResult(component_trace, verbose=False, serialize=None):
     print(str(result))
     return
 
+  elif value_types.HasCustomRepr(result):
+    # Same as above, but for __repr__.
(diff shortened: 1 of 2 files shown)


Borda (approved):

dbieber:
Thanks for the PR @paul-ada! It may be some time before we process it, but it is already appreciated.

PR #442, opened 2023-03-05
Add missing argument description

sp1thas:
This is really minor, this PR aims to provide a description of the `serialize` keyword argument in the corresponding doc-string. In case the wording is not the proper one, feel free to propose a change.

PR #462, opened 2023-09-06
Add missing argument description

sp1thas:
#442 was accidentally closed

This is really minor, this PR aims to provide a description of the serialize keyword argument in the corresponding doc-string. In case the wording is not the proper one, feel free to propose a change.


dbieber:
Thanks for the docstring fix 👍

PR #472, opened 2023-11-13
Fix missing `$` sign in bash completion

maximehk:
Related issue  https://github.com/google/python-fire/issues/64


maximehk:
This should fix errors users face when using `completion.py` in ZSH

dbieber:
Thank you for the fix.

PR #476, opened 2023-12-23
Docstring description multiline parsing 

### LINKED ISSUE #448: parsing function help with multiline
Can I have flexible line breaks in the help comments?

this is the help doc
```
def func(input_table: str):
        """Plot region mutation info using table generated by `bioat bam mpileup_to_table`.

        :param input_table: Path of table generated by `bioat bam mpileup_to_table`.
        This table should contain base mutation information from a short genome region (no more than 1k nt).

```

printed help I got
```
POSITIONAL ARGUMENTS
    INPUT_TABLE
        Type: str
        Path of table generated by `bioat bam mpileup_to_table`. This table should contain base mutation information from a short genome region (no more than 1k nt).

```

printed help I expacted
```
POSITIONAL ARGUMENTS
    INPUT_TABLE
        Type: str
        Path of table generated by `bioat bam mpileup_to_table`. 
        This table should contain base mutation information from a short genome region (no more than 1k nt).
```

thebadcoder96:
this would potentially close #448 

Please review.

### DIFF
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -269,12 +269,7 @@ def _join_lines(lines, type=None):
         group_lines = []
 
   if group_lines:  # Process the final group.
-    # group_text = ' '.join(group_lines)
-    if type == 'description':
-        group_text = '\n'.join(group_lines)
-    else:
-        group_text = ' '.join(group_lines)
-
+    group_text = '\n'.join(group_lines)
     group_texts.append(group_text)
 
   return '\n\n'.join(group_texts)
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -200,7 +200,7 @@ def parse(docstring):
 
   args.extend([KwargInfo(
       name=arg.name, type=_cast_to_known_type(_join_lines(arg.type.lines)),
-      description=_join_lines(arg.description.lines)) for arg in state.kwargs])
+      description=_join_lines(arg.description.lines, 'description')) for arg in state.kwargs])
 
   return DocstringInfo(
       summary=summary,
@@ -239,7 +239,7 @@ def _is_blank(line):
   return not line or line.isspace()
 
 
-def _join_lines(lines):
+def _join_lines(lines, type=None):
   """Joins lines with the appropriate connective whitespace.
 
   This puts a single space between consecutive lines, unless there's a blank
@@ -253,7 +253,7 @@ def _join_lines(lines):
   # TODO(dbieber): Add parameters for variations in whitespace handling.
   if not lines:
     return None
-
+   
   started = False
   group_texts = []  # Full text of each section.
   group_lines = []  # Lines within the current section.
@@ -269,7 +269,12 @@ def _join_lines(lines):
         group_lines = []
 
   if group_lines:  # Process the final group.
-    group_text = ' '.join(group_lines)
+    # group_text = ' '.join(group_lines)
+    if type == 'description':
+        group_text = '\n'.join(group_lines)
+    else:
+        group_text = ' '.join(group_lines)
+
     group_texts.append(group_text)
 
   return '\n\n'.join(group_texts)
--- a/fire/docstrings.py
+++ b/fire/docstrings.py
@@ -200,7 +200,7 @@ def parse(docstring):
 
   args.extend([KwargInfo(
       name=arg.name, type=_cast_to_known_type(_join_lines(arg.type.lines)),
-      description=_join_lines(arg.description.lines, 'description')) for arg in state.kwargs])
+      description=_join_lines(arg.description.lines)) for arg in state.kwargs])
 
   return DocstringInfo(
       summary=summary,
@@ -239,7 +239,7 @@ def _is_blank(line):
   return not line or line.isspace()
 
 
-def _join_lines(lines, type=None):
+def _join_lines(lines):
   """Joins lines with the appropriate connective whitespace.
 
   This puts a single space between consecutive lines, unless there's a blank
--- a/fire/docstrings_test.py
+++ b/fire/docstrings_test.py
@@ -165,7 +165,7 @@ class DocstringsTest(testutils.BaseTestCase):
                     description='The first parameter.'),
             ArgInfo(name='param2', type='str',
                     description='The second parameter. This has a lot of text, '
-                                'enough to cover two lines.'),
+                                'enough to\ncover two lines.'),
         ],
     )
     self.assertEqual(expected_docstring_info, docstring_info)
@@ -229,7 +229,7 @@ class DocstringsTest(testutils.BaseTestCase):
                     description='The second parameter.'),
         ],
         # TODO(dbieber): Support return type.
-        returns='bool True if successful, False otherwise.',
+        returns='bool\nTrue if successful, False otherwise.',
     )
     self.assertEqual(expected_docstring_info, docstring_info)
 
@@ -257,7 +257,7 @@ class DocstringsTest(testutils.BaseTestCase):
                     description='The first parameter.'),
             ArgInfo(name='param2', type='str',
                     description='The second parameter. This has a lot of text, '
-                                'enough to cover two lines.'),
+                                'enough to cover two\nlines.'),
         ],
     )
     self.assertEqual(expected_docstring_info, docstring_info)


fire/docstrings.py
```diff
@@ -177,13 +177,22 @@ def parse(docstring):
...
+  state.max_line_length = 0
 
   for index, line in enumerate(lines):
     has_next = index + 1 < lines_len
     previous_line = lines[index - 1] if index > 0 else None
     next_line = lines[index + 1] if has_next else None
     line_info = _create_line_info(line, next_line, previous_line)
     _consume_line(line_info, state)
+  
+  if state.line2_length:
```
dbieber, line 194: This looks like its set up only to apply the heuristic to the final arg encountered, rather than to each arg separately.

fire/docstrings_test.py
```diff
@@ -257,7 +257,7 @@ def test_numpy_format_multiline_arg_description(self):
                     description='The first parameter.'),
             ArgInfo(name='param2', type='str',
                     description='The second parameter. This has a lot of text, '
-                                'enough to cover two lines.'),
+                                'enough to cover two\nlines.'),
         ],
     )
```
dbieber, line 262: These changes to the tests LGTM.

Let's add new tests. One situation to cover is if there are multiple args requiring line merging. There are probably others too.

fire/docstrings.py
thebadcoder96: True. I have addressed this in commit [93ff111](https://github.com/google/python-fire/pull/476/commits/93ff1118f10aea460f761662feea4d6b25f24739)

fire/docstrings_test.py
thebadcoder96: Added 5 new test cases

fire/docstrings.py
```diff
@@ -410,6 +419,83 @@ def _consume_google_args_line(line_info, state):
   else:
     if state.current_arg:
       state.current_arg.description.lines.append(split_line[0])
+      state.max_line_length = max(state.max_line_length, len(line_info.line))
+      if line_info.previous.line == state.line1: # check for line2
```
dbieber, line 423: This uses string equality which can mistakenly detect a non-first line as being line 1 if the same line is repeated later. Using line indexes rather than line contents would be more robust.

fire/docstrings.py
```diff
@@ -177,6 +177,12 @@ def parse(docstring):
   state.returns.lines = []
   state.yields.lines = []
   state.raises.lines = []
+  state.line1 = None
```
dbieber, line 180: Since you've written this with top-level state rather than per-arg state, I'm concerned about the possibility of state from the first arg leaking into the processing of the second arg.

thebadcoder96: I do not see how it would leak. The only reason I did this with top-level state is otherwise we would have to repeat the below lines twice after line 405 and line 412. Do you think we should go with it being in the arg states?

dbieber: Here's what I meant by possible leakage:

If e.g. state.line3_first_word_length is set from the first arg, and then we start parsing data from the second arg, then state.line3_first_word_length will still be set with its old value.

dbieber: Yes, my inclination is toward arg state rather than top-level state, but I admit I haven't thought it through fully yet.

fire/docstrings.py
thebadcoder96: Understood. My question is how can we use line indexes rather than line_info? Since most of the code consumes the docstring line by line. 

Cmiiw, the check is only done when the line does not have an arg name. Line1 should be a line with the argname. Do you think that line1 (that has arg name in it) would be repeated in any case?

fire/docstrings.py
thebadcoder96: Oooh I get it now! I did face this issue but I believe that our code should probably reset the values back to `None` if there is no line2 or line3. 

```
        else:
          state.line3_first_word_length = None
      else:
        state.line2_first_word_length = state.line2_length = None
```
above code is from line 432.

fire/docstrings.py
dbieber: One way to get the line index would be to add an additional field to line_info, starting by passing the index in as an argument to _create_line_info at line 191.

> Do you think that line1 (that has arg name in it) would be repeated in any case?

I think it would be exceptionally rare, but want to write robust code as much as possible anyway. One reason for doing this is so that the code is more likely to stay correct when future changes are made.

fire/docstrings.py
thebadcoder96: The `state.max_line_length` would leak the max line length from arg1 to arg2 but I think we want that right?

We can move everything else except the max_line_length to the arg state. Wdyt?

dbieber: Makes sense to me.

We do have to be thoughtful then about whether we want to process the early args using the max_line_length available at the time that arg is ingested, or if we wait to the end to use the same max_line_length for all args.

thebadcoder96: I chose to find the `state.max_line_length` right at the start (max line length of the whole docstring) and have updated the code to be per-arg state.

thebadcoder96:
Sorry I was slacking off a bit :P 

I have found another issue with the code (a new possible test case). When I use `:` in the second or third line, the line disappears since it is detected as an arg with the current logic. I also found that my line3 logic to check if it was an arg was wrong. 

Ex: if I had something like this:
```
    param3_that_is_very_longer_than_usual: The second parameter. This has a lot of text,
      enough to: cover two lines.
```
it would print out:
```
PARAM3_THAT_IS_VERY_LONGER_THAN_USUAL
        The second parameter. This has a lot of text, 
```

The problem is with `_as_arg_name_and_type()` because when the part split by `:` passes through it, it only checks if there is more than 1 string split by ' ' and if the first word looks like an arg, combines everything else into a string removing `()[]{}`. I have added another condition to check if these brackets are present to qualify the whole string as an arg and type. 

When I made this change I noticed that around line 408/417
```
      else:
        if state.current_arg:
          state.current_arg.description.lines.append(split_line[0])
```
here we are appending only `split_line[0]` but we should be joining everything back with `:` and then appending it. I made this change as well. Then I realized that we never do the check for line2 and line3 and store their values here so if there is a line with `:` that is not an arg it does not check for the code I have added. 

I decided to convert the check into a function and call them in both places when the line is not an arg. Added 3 more tests regarding the `:`

fire/docstrings.py
thebadcoder96: I have added the index as you guided and updated the code to compare indexes instead of strings. Please review the most recent commit.

dbieber:
Ack. That all sounds good. Will update you when I get to the review.

thebadcoder96:
> Ack. That all sounds good. Will update you when I get to the review.

What I mentioned might be another bug. I also just thought about this; I was messing around adding `:` in the second and third lines of the description. I always added it after 2 words (like in the example mentioned above) which caused me to believe `_as_arg_name_and_type ` was the problem.

I haven't tested this, but if we add `:` after the first word in descriptions (line2 or later), the `_is_arg_name()` function will be called and according to our current logic, anything that is just a word will be considered as an arg. I think we want a better way to identify an arg. 

On the top of my head, I'm thinking we can do something like compare the previous line if `state.current_arg` exists and see if there is a difference in indentation (most likely not an arg if line2 first word and arg name indentation difference is present). Just spit-balling my thoughts right now.

fire/docstrings_test.py
```diff
@@ -165,7 +165,208 @@ def test_google_format_multiline_arg_description(self):
...
+    expected_docstring_info = DocstringInfo(
+        summary='This is a Google-style docstring with long args.',
+        args=[
+            ArgInfo(name='function_maker_handler_event_factory',
+                    description='The function-maker-handler event factory '
+        'responsible for making the function-maker-handler event. '
+        'Use this whenever\nyou need a function-maker-handler event'
+         ' made for you programmatically.'),
+        ],
+    )
```
jamesowers-cohere, line 191: It may be good to add an example here, or below, which contains a second line in the documentation before the args (since this is very common). e.g. 
```suggestion
    docstring = """This is a Google-style docstring with long args.
    
    Here is some further and potentially long-winded explanation of the function.
    
    Args:
        function_maker_handler_event_factory: The function-maker-handler event factory
            responsible for making the function-maker-handler event. Use this whenever
            you need a function-maker-handler event made for you programmatically. 
    """
    docstring_info = docstrings.parse(docstring)
    expected_docstring_info = DocstringInfo(
        summary='This is a Google-style docstring with long args.\n\nHere is some further and potentially long-winded explanation of the function.',
        args=[
            ArgInfo(name='function_maker_handler_event_factory',
                    description='The function-maker-handler event factory '
        'responsible for making the function-maker-handler event. '
        'Use this whenever\nyou need a function-maker-handler event'
         ' made for you programmatically.'),
        ],
    )
```

thebadcoder96: I like it. I think this test case would fail rn since the default behavior is to ignore the \n and just display without them. 

This was the issue I fixed but I think I have fixed it within the scope of args but not the name of the function and summary. I will work on this as well.

thebadcoder96: Another thing I noticed was that the name of the function is displayed as <name> -  <summary>. Is that how it should be? I only ask since we have a section called description where we display the summary.

dbieber: > Another thing I noticed was that the name of the function is displayed as <name> -  <summary>.

I think so, though maybe there are examples where it doesn't make sense and we can improve it. I'd have to give that some thought, but fortunately that's orthogonal to this change.

fire/docstrings.py
```diff
@@ -336,9 +342,10 @@ def _as_arg_name_and_type(text):
     None otherwise.
   """
   tokens = text.split()
+  is_type = any(c in "[](){}" for c in text)
```
dbieber, line 345: What was the reason for this line?

thebadcoder96: > I haven't tested this, but if we add : after the first word in descriptions (line2 or later), the _is_arg_name() function will be called and according to our current logic, anything that is just a word will be considered as an arg. I think we want a better way to identify an arg.

This was the initial reason for this. I just came up with this for the time being but i think that we need a better way to identify an arg name.

PR #490, opened 2024-03-11
ci: expand build matrix for major OS

Borda:
I suggest including testing on Mac and Win. 
also, the addition of OS builds does not need to be so extensive and could be listed in the `include` section adding 3.8 as the oldest and 3.12 as the latest python version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,12 +2,22 @@ name: Python Fire
 
 on: [push]
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.5", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        fail-fast: false
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        include:
+          # adding some specific/older versions to the matrix
+          - {os: "ubuntu-20.04", python-version: "3.5"}
+          - {os: "ubuntu-20.04", python-version: "3.7"}
 
     steps:
      # Checkout the repo.
@@ -22,7 +32,6 @@ jobs:
 
      # Build Python Fire using the build.sh script.
      - name: Run build script
-       shell: bash
        run: ./.github/scripts/build.sh
        env:
          PYTHON_VERSION: ${{ matrix.python-version }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -2,12 +2,22 @@ name: Python Fire
...
+    shell: bash
+
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.5", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        fail-fast: false
```
Borda, line 14:
```suggestion
        fail-fast: true
```

hugovk: I recommend `false`, this way you get to see what exactly failed -- is it all versions lower than 3.10? Is it Windows?

With `true`, if just the 3.5/Windows job failed, and all the others were cancelled, you don't get this useful info.

.github/workflows/build.yml
```diff
@@ -2,12 +2,22 @@ name: Python Fire
...
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
```
hugovk, line 15: Starting slower ones first can prevent bottlenecks and help builds as a whole finish sooner.

```suggestion
        os: ["windows-latest", "macos-latest", "ubuntu-latest"]
```

.github/workflows/build.yml
Borda: true is by default, just make it explicit, I normally prefer to run all but it could yield quite extensive testing

Borda:
@dbieber friendly ping if you could have a look... :chipmunk:

dbieber:
I'm open to this!

Borda:
> I'm open to this!

@dbieber updated and resolved conflicts

dbieber:
I'm seeing this error
```
[Invalid workflow file: .github/workflows/build.yml#L18](https://github.com/google/python-fire/actions/runs/10973863694/workflow)
The workflow is not valid. .github/workflows/build.yml (Line: 18, Col: 20): Unexpected value 'false'
```

.github/workflows/build.yml
Borda:
```suggestion
```

Borda:
> I'm seeing this error

Let's go with @hugovk

dbieber:
> @dbieber if I won't need approval for each run, it would be great (faster)

It should run without approval now. 👍

.github/workflows/build.yml
```diff
@@ -2,18 +2,23 @@ name: Python Fire
...
+    shell: bash
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13.0-rc.2"]
+        os: ["windows-latest", "macos-latest", "ubuntu-latest"]
```
Borda, line 18:
```suggestion
        os: ["macos-latest", "ubuntu-latest"]
```
Let's skip win until the build script is updated

PR #503, opened 2024-03-22
replaced sum of strings for fstrings

igeni:
replaced sum of strings for fstrings to avoid potential type's mismatch and make code more clear

### DIFF
--- a/fire/helptext.py
+++ b/fire/helptext.py
@@ -111,7 +111,7 @@ def _NameSection(component, info, trace=None, verbose=False):
                                              LINE_LENGTH)
 
   if summary:
-    text = current_command + ' - ' + summary
+    text = f'{current_command} - {summary}'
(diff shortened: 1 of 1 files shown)


fire/helptext.py
```diff
@@ -582,7 +582,7 @@ def _CreateItem(name, description, indent=2):
 def _GetArgDescription(name, docstring_info):
   if docstring_info.args:
     for arg_in_docstring in docstring_info.args:
-      if arg_in_docstring.name in (name, '*' + name, '**' + name):
+      if arg_in_docstring.name in (f'{name}', f'*{name}', f'**{name}'):
```
dbieber, line 585: f'{name}' should just be name

dbieber:
These changes are included in https://github.com/google/python-fire/pull/551

PR #513, opened 2024-05-21
bug fix for help from typing package, add test for optional and union

### LINKED ISSUE #508: Optional type args are mistyped in help
Type annotations that use `typing.Optional` are incorrectly documented when showing help.

## Example 1

```python
from typing import Optional

import fire

def main(arg: Optional[str] = 'something'):
    """Do something with arg."""


if __name__ == '__main__':
    fire.Fire(main)
```

The corresponding help output is:

```
INFO: Showing help with the command 'firetest.py -- --help'.

NAME
    firetest.py - Do something with arg.

SYNOPSIS
    firetest.py <flags>

DESCRIPTION
    Do something with arg.

FLAGS
    -a, --arg=ARG
        Type: Optional
        Default: 'something'
```

**Here, `Type: Optional` is totally incorrect. It should be `Optional[str]`.** My use of `str` is just an example and it can be anything.

## Example 2

```python
from typing import Optional

import fire


def main(arg: Optional[str] = None):
    """Do something with arg."""


if __name__ == '__main__':
    fire.Fire(main)
```

The corresponding help output is:

```
INFO: Showing help with the command 'firetest.py -- --help'.

NAME
    firetest.py - Do something with arg.

SYNOPSIS
    firetest.py <flags>

DESCRIPTION
    Do something with arg.

FLAGS
    -a, --arg=ARG
        Type: Optional[Optional]
        Default: None
```

**Here, `Type: Optional[Optional]` is even more incorrect. It should be `Optional[str]`.** My use of `str` is just an example and it can be anything.

Jemeljanov:
Closes #508 

Fixed bug where objects from typing instance were not displayed correctly mentioned in #508 and also Union types.

Added tests for typing.Union and typing.Optional

### Other notes
Removed comment as the edge case mentioned is handled by new code.

Added a regular expression to output a shorter message, also not sure that a longer message would be beneficial to user.

### DIFF
--- a/fire/helptext_test.py
+++ b/fire/helptext_test.py
@@ -167,6 +167,42 @@ class HelpTest(testutils.BaseTestCase):
   @testutils.skipIf(
       sys.version_info[0:2] < (3, 5),
       'Python < 3.5 does not support type hints.')
+  def testHelpTextFunctionWithTypesAndDefaultNoneFromTypingOptional(self):
+    component = (
+        tc.py3.WithDefaultsAndTypes().typing_optional_get_int)  # pytype: disable=module-attr
+    help_screen = helptext.HelpText(
+        component=component,
+        trace=trace.FireTrace(component, name='get_int'))
+    self.assertIn('NAME\n    get_int', help_screen)
+    self.assertIn('SYNOPSIS\n    get_int <flags>', help_screen)
+    self.assertNotIn('DESCRIPTION', help_screen)
+    self.assertIn(
+        'FLAGS\n    -v, --value=VALUE\n'
+        '        Type: Optional[int]\n        Default: None',
+        help_screen)
+    self.assertNotIn('NOTES', help_screen)
+
+  @testutils.skipIf(
+      sys.version_info[0:2] < (3, 5),
+      'Python < 3.5 does not support type hints.')
+  def testHelpTextFunctionWithTypesAndDefaultNoneFromTypingUnion(self):
+    component = (
+        tc.py3.WithDefaultsAndTypes().typing_union_get_int)  # pytype: disable=module-attr
+    help_screen = helptext.HelpText(
+        component=component,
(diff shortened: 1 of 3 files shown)


fire/helptext_test.py
```diff
@@ -164,6 +164,42 @@ def testHelpTextFunctionWithTypesAndDefaultNone(self):
...
+        tc.py3.WithDefaultsAndTypes().typing_union_get_int)  # pytype: disable=module-attr
+    help_screen = helptext.HelpText(
+        component=component,
+        trace=trace.FireTrace(component, name='get_int'))
+    self.assertIn('NAME\n    get_int', help_screen)
+    self.assertIn('SYNOPSIS\n    get_int <flags>', help_screen)
+    self.assertNotIn('DESCRIPTION', help_screen)
+    self.assertIn(
+        'FLAGS\n    -v, --value=VALUE\n'
+        '        Type: Optional[int, str]\n        Default: None',
```
dbieber, line 199: This should read Union[int, str], yes?

dbieber:
Thanks for putting this together! At a quick glance, this LGTM. One comment inline above.

fire/helptext_test.py
Jemeljanov: It displays Optional because it has a default value.

Another example for:
`def main(arg: Union[str, int]):
    pass` 
The output is 

`    ARG
        Type: str, int`

Should test be added also for this?

Jemeljanov:
@dbieber it was a little bit different behaviour.

Now it is more in compliance with the typing in python code and a more robust version than with regular expression.

Additional check is to prevent the Optional repetition (one added from python-fire the other from typing package).

dbieber:
I only just enabled CI for pull requests so this didn't surface earlier -- it looks like there are some test failures.

Borda (approved):

PR #526, opened 2024-07-22
Use ast.Constant for recent Python versions

### LINKED ISSUE #523: Replace deprecated usage of ast.Str
## Motivation

This tool depends on `ast.Str` to wrap any value other than `True`, `False` and `None`,
which is going to be removed in Python 3.14 (probably in less than one year).

Before the day that Python 3.14 is released, it would be great to change the usage
of `ast.Str` to `ast.Constant` (added in Python 3.8), and make a new release.
With this change, we can make sure the `fire` library still works after Python 3.14 is released.

## Possible Solution

It could be solve quickly by one-line change, although we need conditional import
to preserve backward compatibility.

```diff
diff --git a/fire/parser.py b/fire/parser.py
index 2aff8bd..bdf3cdb 100644
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -20,7 +20,12 @@ from __future__ import print_function
 
 import argparse
 import ast
+import sys
 
+if sys.version_info[0:2] < (3, 8):
+  _StrNode = ast.Str
+else:
+  _StrNode = ast.Constant
 
 def CreateParser():
   parser = argparse.ArgumentParser(add_help=False)
@@ -127,4 +132,4 @@ def _Replacement(node):
   # These are the only builtin constants supported by literal_eval.
   if value in ('True', 'False', 'None'):
     return node
-  return ast.Str(value)
+  return _StrNode(value)
```

All existing test cases passed.

If this change looks reasonable,
I could submit a PR for this, saving the time for maintainer. :D

## References

- [What's New in Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#deprecated)
- [What's New in Python 3.14](https://docs.python.org/3.14/whatsnew/3.14.html#removed)

wdhongtw:
ast.Str is planned to removed in Python 3.14, use ast.Constant instead whenever the later is available.

This PR resolves #523, if that's an issue. :joy_cat: 

All test cases are passed and the pylint score is not changed.

> Your code has been rated at 9.87/10 (previous run: 9.87/10, +0.00)

### DIFF
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -20,7 +20,12 @@ from __future__ import print_function
 
 import argparse
 import ast
+import sys
 
+if sys.version_info[0:2] < (3, 8):
+  _StrNode = ast.Str
+else:
+  _StrNode = ast.Constant
 
 def CreateParser():
   parser = argparse.ArgumentParser(add_help=False)
@@ -127,4 +132,4 @@ def _Replacement(node):
   # These are the only builtin constants supported by literal_eval.
   if value in ('True', 'False', 'None'):
     return node
-  return ast.Str(value)
+  return _StrNode(value)


dbieber:
Thanks for the fix!

PR #549, opened 2024-09-20
Upgrade pylint version

dbieber:


Borda:
how about switching from pylint to [Ruff](https://docs.astral.sh/ruff/) as so many major projects?
if you are fine with it I can do it...

dbieber:
Yes, I would be on board with adopting uv and ruff.

PR #557, opened 2024-09-21
Parallelize pytype

dbieber:


dbieber:
Slower than before. Closing without merging.

PR #560, opened 2024-09-23
lint: configure pre-commit with Ruff

Borda:
reaction to https://github.com/google/python-fire/pull/549#issuecomment-2365220214

Switch from `pylint` to Ruff with the same rules to be applied
the Ruff is implemented as pre-commit hook for simplicity


dbieber:
Ah, we might actually need all the "# pylint: " directives for the google-internal fire codebase. Let's hold off on this.

Borda:
> Ah, we might actually need all the "# pylint: " directives for the google-internal fire codebase. Let's hold off on this.

good to know, in such case would nee to keep pylint... :(

Borda:
@dbieber friendly ping if we shall land it or close it? :)

PR #609, opened 2025-06-01
Bump version to 0.7.1

dbieber:


qthequartermasterman:
It would be great to have a new release made, especially since there are fixes (like an incompatibility with newer versions of ipython) on master that haven't been released yet. I am having to pin ipython to use this package.

PR #611, opened 2025-06-17
Visualziations of the python-fire codebase

ivanmilevtues:
This PR contains high-level diagrams for the python-fire codebase. You can see how they render in Github here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/python-fire/on_boarding.md

These diagrams help users quickly grasp the codebase. Google is a huge organization so I would love to hear what is your current on-boarding process, when it comes to working with new codebases. Do you think that such diagrams can have place in it? We are generating the with static analysis and LLMs, so they are based on the code itself without external resources.

Any feedback is more than welcome! We've just released a free GitHub Action that can automatically update the diagrams and this way keep them always up-to-date.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Argument_Input_Processing.md
+++ b/.codeboarding/Argument_Input_Processing.md
@@ -1,0 +1,71 @@
+```mermaid
+
+graph LR
+
+    Argument_Input_Processing["Argument & Input Processing"]
+
+    Core_CLI_Engine["Core CLI Engine"]
+
+    Argument_Input_Processing -- "provides processed arguments to" --> Core_CLI_Engine
+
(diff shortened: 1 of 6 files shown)


dbieber:
Thanks for offering, but I don't plan on incorporating these diagrams at this time.

ivanmilevtues:
Hey @dbieber, just dropping by to let you know that now we are open source: https://github.com/CodeBoarding/CodeBoarding
So if you are interested you can take a look at the codebase :)

Issue #382, opened 2022-02-01
BUG: When using a Class for grouping commands `--self` shows as a flag


chey:
### Example code (example.py):
```python
import fire

class Foo:
    def bar(self):
        return 'bar'

    def do_stuff(self, arg1):
        return 'stuff'

if __name__ == '__main__':
    fire.Fire(Foo)
```

### Run:
`python3 example.py -- --completion`

### Completion Output (take not of `--self` in output):
```bash
# bash completion support for example.py
# DO NOT EDIT.
# This script is autogenerated by fire/completion.py.

_complete-examplepy()
{
  local cur prev opts lastcommand
  COMPREPLY=()
  prev="${COMP_WORDS[COMP_CWORD-1]}"
  cur="${COMP_WORDS[COMP_CWORD]}"
  lastcommand=$(get_lastcommand)

  opts=""
  GLOBAL_OPTIONS=""


  case "${lastcommand}" in

    bar)

      if is_prev_global; then
        opts="${GLOBAL_OPTIONS}"
      else
        opts="--self ${GLOBAL_OPTIONS}"
      fi
      opts=$(filter_options $opts)
    ;;

    do-stuff)

      if is_prev_global; then
        opts="${GLOBAL_OPTIONS}"
      else
        opts="--arg1 --self ${GLOBAL_OPTIONS}"
      fi
      opts=$(filter_options $opts)
    ;;

    example.py)

      opts="bar do-stuff ${GLOBAL_OPTIONS}"
      opts=$(filter_options $opts)
    ;;
  esac

  COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
  return 0
}

get_lastcommand()
{
  local lastcommand i

  lastcommand=
  for ((i=0; i < ${#COMP_WORDS[@]}; ++i)); do
    if [[ ${COMP_WORDS[i]} != -* ]] && [[ -n ${COMP_WORDS[i]} ]] && [[
      ${COMP_WORDS[i]} != $cur ]]; then
      lastcommand=${COMP_WORDS[i]}
    fi
  done

  echo $lastcommand
}

filter_options()
{
  local opts
  opts=""
  for opt in "$@"
  do
    if ! option_already_entered $opt; then
      opts="$opts $opt"
    fi
  done

  echo $opts
}

option_already_entered()
{
  local opt
  for opt in ${COMP_WORDS[@]:0:COMP_CWORD}
  do
    if [ $1 == $opt ]; then
      return 0
    fi
  done
[...]

PR #650, opened 2026-03-08
Fix completion suggesting --self for class-grouped commands

shivamtiwari3:
Fixes google/python-fire issue #382: https://github.com/google/python-fire/issues/382

What happened
When using a class as the Fire component (to group commands), Fire’s --completion output could include the implicit bound argument self as a CLI flag (--self). This is confusing and incorrect since users can never pass self explicitly.

Example symptom (from completion script output):

opts="--arg1 --self ${GLOBAL_OPTIONS}"
opts="--self ${GLOBAL_OPTIONS}"
Root cause
The completion generator builds option flags from the callable’s argspec:

Completions() → inspectutils.GetFullArgSpec(component) → _CompletionsFromArgs(spec.args + spec.kwonlyargs)
For unbound methods (functions retrieved from a class), spec.args includes self (or cls), because Python includes that parameter in the signature even though it is implicitly supplied at call time.
_CompletionsFromArgs was converting every arg name into a --<arg> completion, which incorrectly produced --self.

Fix
In fire/completion.py, _CompletionsFromArgs now skips a leading self / cls argument when present. This ensures completion suggestions only contain user-supplied parameters.

This is a minimal, targeted change that affects only completion flag generation from argument lists.

Reproduction
A standalone repro script is included:

python3 repro_issue_382.py
It generates completion output for a small example class and asserts that --self is not present in the completion script.

Tests
python3 -m unittest fire.completion_test -q
Notes
This change is limited to completion generation. It does not affect Fire’s argument parsing or invocation behavior.
Existing test testMethodCompletions already asserts that --self should not appear; this change ensures the completion script output also respects that constraint in the class-grouping scenario.

### DIFF
--- a/fire/completion.py
+++ b/fire/completion.py
@@ -379,6 +379,14 @@ def _CompletionsFromArgs(fn_args):
   Returns:
     A list of possible completion strings for that function.
   """
+  # Completion should not suggest implicit bound args.
+  #
+  # When generating completions, Fire may inspect unbound methods (functions
+  # retrieved from a class). These will include "self" or "cls" in the argspec,
+  # even though users can never pass those via the CLI.
+  if fn_args and fn_args[0] in ('self', 'cls'):
+    fn_args = fn_args[1:]
+
   completions = []
   for arg in fn_args:
     arg = arg.replace('_', '-')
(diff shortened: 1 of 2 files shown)


shivamtiwari3:

PR #651, opened 2026-03-09
Use Python's logging framework for INFO diagnostic messages (fixes #353)

### LINKED ISSUE #353: Logging messages don't using logging
https://github.com/google/python-fire/blob/c1266d0dbb2114514fcf8be62044344b5a51c733/fire/core.py#L239
https://github.com/google/python-fire/blob/c1266d0dbb2114514fcf8be62044344b5a51c733/fire/core.py#L287

When I get the help message, includes INFO: messages like:
    INFO: Showing help with the command {cmd}

I would like such logging messages to use Python's logging framework which would allow me to suppress and redirect such messages.

shivamtiwari3:
## Summary

Fixes #353.

The two `INFO: Showing help …` messages that Python Fire prints to stderr
were emitted via bare `print()` calls.  This made it impossible for users
to suppress or redirect the messages using Python's standard `logging`
machinery.

---

## Problem

```python
import fire

# There was no way to suppress this line before this fix:
# INFO: Showing help with the command 'mytool -- --help'.
fire.Fire(MyClass, command=['--help'])
```

A common workaround was the fragile monkey-patch described in #353:

```python
def _print(*args, **kwargs):
    args = [a for a in args if not (isinstance(a, str) and a.startswith('INFO:'))]
    if args:
        print(*args, **kwargs)
fire.core.print = _print
```

---

## Solution

Three changes in `fire/core.py`:

### 1. `_LazyStderrStreamHandler`

A thin `logging.StreamHandler` subclass that resolves `sys.stderr`
**at emit time** via a `@property` rather than capturing it at
construction time.  This is essential for correctness in tests that
replace `sys.stderr` with an in-memory `StringIO` buffer using
`unittest.mock.patch.object(sys, 'stderr', ...)`.

### 2. `_logger`

```python
_logger = logging.getLogger(__name__)   # → 'fire.core'
```

The handler is added only when `_logger` has no handlers yet (idempotent
on repeated imports / reloads).  `propagate` is set to `False` so that
users who configure the root logger do not receive duplicate output.

### 3. Level on the parent `'fire'` logger

The level is set to `INFO` on the `'fire'` parent logger (not on
`'fire.core'` directly).  Because `'fire.core'` inherits its *effective*
level from the hierarchy, users can suppress **all** fire messages with a
single call to the parent:

```python
import logging
logging.getLogger('fire').setLevel(logging.WARNING)   # suppress INFO
```

or target just this module:

```python
logging.getLogger('fire.core').setLevel(logging.WARNING)
```

or redirect to a custom destination:

```python
handler = logging.FileHandler('fire.log')
logging.getLogger('fire').handlers = [handler]
```

### Default behaviour is unchanged

For a user who does **not** configure logging, the `INFO:` line continues
to appear on stderr exactly as before.  The format is preserved:

```
INFO: Showing help with the command 'mytool -- --help'.
```

---

## Tests added (`LoggingTest` in `fire/core_test.py`)

| Test | What it verifies |
|---|---|
| `testInfoMessageAppearsOnStderrByDefault` | Default stderr output is preserved (no regression) |
| `testInfoMessageUsesFireLogger` | `core._logger` is a `logging.Logger` named `'fire.core'` |
| `testInfoCanBeSuppressedViaLogging` | Setting `fire.core` level to `WARNING` hides the INFO line |
| `testInfoCanBeSuppressedViaParentLogger` | Setting `fire` level to `WARNING` hides the INFO line |
| `testInfoCanBeRedirectedViaCustomHandler` | Replacing handlers redirects INFO to a custom stream |

---

## Test results

```
265 passed in 0.25s
```

All pre-existing tests pass without modification.  The four `INFO:.*SYNOPSIS`
assertions in `CoreTest` continue to match because the logging output format
(`INFO: <message>`) is identical to what `print()` produced before.

---

## Checklist

- [x] Fixes the reported issue (#353)
- [x] Default user-visible behaviour is unchanged
- [x] Uses `logging.getLogger(__name__)` (library best practice)
- [x] No new runtime dependencies
- [x] Fully covered by new and existing tests
- [x] All 265 tests pass

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -52,6 +52,7 @@ The available flags for all Fire CLIs are:
 import asyncio
 import inspect
 import json
+import logging
 import os
 import re
 import shlex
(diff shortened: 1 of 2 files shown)


shivamtiwari3:

PR #652, opened 2026-03-10
Fix short flags being swallowed by **kwargs (fixes #454)

### LINKED ISSUE #454: [bug] short arguments does not work with `**kwargs`
# Description:

Short arguments (e.g. `-u` for `--username`) does not work with `**kwargs`

# Example that works OK:

The following example works ok:

```python
import fire

def cli(first_arg="left", second_arg="right"):
    """Prints the 2 arguments separated by bar
       
    Parameters
    ----------
    first_arg : str
        the first arg
    second_arg : str
        the second arg
    """
    print(f"{first_arg} | {second_arg}" )

if __name__ == '__main__':
  fire.Fire(cli)
```
## Help output
```sh
❯ python .\example.py -- --help
NAME
    example.py - Prints the 2 arguments separated by bar

SYNOPSIS
    example.py <flags>

DESCRIPTION
    Prints the 2 arguments separated by bar

FLAGS
    -f, --first_arg=FIRST_ARG
        Default: 'left'
        the first arg
    -s, --second_arg=SECOND_ARG
        Default: 'right'
        the second arg
```
## Usage with short flags
```sh
❯ python .\example.py -f="hello" -s="word"
hello | word
```
# Example that does not work:

The following next example does not work with short flags, **even when they are showing on the help output**.
It is using `**kwargs`. It seems the short flags going to the `**kwargs` dictionary.

```python
import fire

def cli(first_arg="left", second_arg="right", **kwargs):
    """Prints the 2 arguments separated by bar
       
    Parameters
    ----------
    first_arg : str
        the first arg
    second_arg : str
        the second arg
    kwargs : str
        additional args
    """
    print(f"{first_arg} | {second_arg}" )

if __name__ == '__main__':
  fire.Fire(cli)
```
## Help output
```sh
❯ python .\example2.py -- --help
NAME
    example2.py - Prints the 2 arguments separated by bar

SYNOPSIS
    example2.py <flags>

DESCRIPTION
    Prints the 2 arguments separated by bar

FLAGS
    -f, --first_arg=FIRST_ARG
        Default: 'left'
        the first arg
    -s, --second_arg=SECOND_ARG
        Default: 'right'
        the second arg
    Additional flags are accepted.
        additional args
```
## Usage with short flags
```sh
❯ python .\example2.py -f="hello" -s="word"
left | right
{'f': 'hello', 's': 'word'}
```

shivamtiwari3:
## Summary

Fixes #454.

Short flags like `-f` and `-s` were being stored directly in `**kwargs` (as `{'f': 'hello', 's': 'world'}`) instead of being expanded to their matching positional argument names (`first_arg`, `second_arg`) when a function accepts `**kwargs`.

---

## Problem

In `_ParseKeywordArgs` ([core.py:887](https://github.com/google/python-fire/blob/master/fire/core.py#L887)), the condition was:

```python
if (key in fn_args
    or (is_bool_syntax and key.startswith('no') and key[2:] in fn_args)
    or fn_keywords):
  keyword = key
elif len(key) == 1:
  # single-char shortcut lookup ...
```

When `fn_keywords` is truthy (i.e. the function has `**kwargs`), the first branch fires immediately, setting `keyword = key` (the raw single character). The shortcut lookup in the `elif` branch is never reached.

---

## Solution

Restructured the conditions so single-char shortcut expansion against `fn_args` is always attempted first. `fn_keywords` is only used as a fallback when no `fn_arg` matches the single character:

```python
if key in fn_args or (...noXxx... in fn_args):
  keyword = key
elif len(key) == 1:
  matching_fn_args = [arg for arg in fn_args if arg[0] == key]
  if len(matching_fn_args) == 1:
    keyword = matching_fn_args[0]
  elif len(matching_fn_args) > 1:
    raise FireError(...)
  elif fn_keywords:      # ← fallback: no fn_arg match, accept into **kwargs
    keyword = key
elif fn_keywords:
  keyword = key
```

Multi-character unrecognised flags still fall through to `**kwargs` as before.

---

## Testing

- Added `fn_with_defaults_and_kwargs` to `test_components.py`
- Added `testSingleCharFlagParsingWithKwargs` to `fire_test.py` covering:
  - `-f=hello -s=world` → `('hello', 'world', {})`
  - `-f hello -s world` → `('hello', 'world', {})`
  - unrecognised `--unknown_key=42` → still forwarded to `**kwargs`
- Command: `python3 -m pytest fire/ --ignore=fire/docstrings_fuzz_test.py --ignore=fire/parser_fuzz_test.py`
- **261 tests pass, 0 failures**

---

## Checklist

- [x] Fixes the reported issue
- [x] Existing behaviour unchanged (no regressions)
- [x] Tests added
- [x] All 261 tests pass
- [x] Code style matches project conventions

### DIFF
--- a/fire/core.py
+++ b/fire/core.py
@@ -884,12 +884,13 @@ def _ParseKeywordArgs(args, fn_spec):
 
       # Determine the keyword.
       keyword = ''  # Indicates no valid keyword has been found yet.
-      if (key in fn_args
-          or (is_bool_syntax and key.startswith('no') and key[2:] in fn_args)
-          or fn_keywords):
+      if key in fn_args or (
+          is_bool_syntax and key.startswith('no') and key[2:] in fn_args):
         keyword = key
       elif len(key) == 1:
-        # This may be a shortcut flag.
+        # This may be a shortcut flag. Check fn_args before falling back to
+        # fn_keywords so that short flags like -f expand to the matching
+        # fn_arg (e.g. 'first_arg') even when **kwargs is present.
         matching_fn_args = [arg for arg in fn_args if arg[0] == key]
         if len(matching_fn_args) == 1:
           keyword = matching_fn_args[0]
@@ -898,6 +899,10 @@ def _ParseKeywordArgs(args, fn_spec):
               f"The argument '{argument}' is ambiguous as it could "
               f"refer to any of the following arguments: {matching_fn_args}"
           )
+        elif fn_keywords:
+          keyword = key
+      elif fn_keywords:
+        keyword = key
 
       # Determine the value.
       if not keyword:
--- a/fire/fire_test.py
+++ b/fire/fire_test.py
@@ -491,6 +491,23 @@ class FireTest(testutils.BaseTestCase):
         fire.Fire(tc.SimilarArgNames,
                   command=['identity2', '-a', '-alpha']), (True, True))
 
+  def testSingleCharFlagParsingWithKwargs(self):
+    # Short flags should map to their matching fn_arg even when **kwargs is
+    # present (regression test for github.com/google/python-fire/issues/454).
+    self.assertEqual(
+        fire.Fire(tc.fn_with_defaults_and_kwargs,
+                  command=['-f=hello', '-s=world']),
+        ('hello', 'world', {}))
+    self.assertEqual(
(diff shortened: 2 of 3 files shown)


shivamtiwari3:

PR #653, opened 2026-03-10
Fix HasCustomStr to recognise custom __repr__ as a meaningful str (fixes #595)

### LINKED ISSUE #595: Fire can't find custom `__str__` of `~datasets.Dataset` from HuggingFace
Hi,

I am using fire on a `main` function that returns a `~datasets.Dataset` object. When running from cli, I get the help screen for `~datasets.Dataset` rather than the `__str__` implementation (likely inherited from a parent class -- haven't looked).

## Reproducible example

```python
from datasets import Dataset
import fire

def main():
    # Create a simple dataset
    dataset = Dataset.from_dict({"a": [1, 2, 3], "b": [4, 5, 6]})
    return dataset

if __name__ == "__main__":
    fire.Fire(main)
    
    # This works but I feel shouldn't be necessary:
    # def custom_serializer(obj: object) -> str:
    #     if hasattr(obj, "__str__"):
    #         return str(obj)
    #     else:
    #         return ""
    # fire.Fire(main, serialize=custom_serializer)
```

shivamtiwari3:
## Summary

Fixes #595.

Objects that override `__repr__` but not `__str__` (e.g. `datasets.Dataset` from HuggingFace) were shown as a help screen instead of being printed, because `HasCustomStr` only checked for a custom `__str__`.

---

## Problem

Python's `object.__str__` delegates to `__repr__`, so a class that defines a custom `__repr__` already produces a meaningful custom string via `str(obj)` — even without defining `__str__`. However, `HasCustomStr` in `value_types.py` only inspected `__str__`:

```python
str_attr = class_attrs.get('__str__')
if str_attr and str_attr.defining_class is not object:
    return True
# __repr__-only classes fall through → False → help screen shown
```

This caused `fire.Fire(main)` to show the `datasets.Dataset` help screen rather than the dataset's repr string.

---

## Solution

After the existing `__str__` check, also look for a custom `__repr__` — but only on **user-defined (non-stdlib) types**, so that Python built-in containers (`dict`, `list`, `OrderedDict`, …) continue to be treated as navigable groups rather than values.

The guard uses `sys.stdlib_module_names` (Python ≥ 3.10) and falls back to a minimal sentinel set for older interpreters:

```python
_STDLIB_MODULE_NAMES = frozenset(
    getattr(sys, 'stdlib_module_names', {'builtins', 'collections', 'abc'})
)
```

If the `__repr__`-defining class's top-level module is not in that set, the object is treated as a value.

---

## Testing

Added to `test_components.py`:
- `ClassWithCustomRepr` — overrides `__repr__` only
- `ClassWithCustomStr`  — overrides `__str__` only (existing behaviour)
- `fn_returning_custom_repr` / `fn_returning_custom_str` — functions returning those objects

Added to `core_test.py`:
- `testPrintObjectWithCustomStr` — existing `__str__` path still works
- `testPrintObjectWithCustomRepr` — new `__repr__` path: output is `ClassWithCustomRepr()`, not a help screen

Run command: `python3 -m pytest fire/ --ignore=fire/docstrings_fuzz_test.py --ignore=fire/parser_fuzz_test.py`
**262 tests pass, 0 failures.**

---

## Checklist

- [x] Fixes the reported issue
- [x] Existing behaviour unchanged (no regressions — dict/list/OrderedDict still show help screens)
- [x] Tests added
- [x] All 262 tests pass
- [x] Code style matches project conventions

### DIFF
--- a/fire/core_test.py
+++ b/fire/core_test.py
@@ -161,6 +161,19 @@ class CoreTest(testutils.BaseTestCase):
     with self.assertOutputMatches(stdout='11', stderr=None):
       core.Fire(tc.NamedTuple, command=['point', '-2'])
 
+  def testPrintObjectWithCustomStr(self):
+    # An object that overrides __str__ should be printed as a value, not shown
+    # as a help screen (regression test for github.com/google/python-fire#595).
+    with self.assertOutputMatches(stdout='custom str', stderr=None):
+      core.Fire(tc.fn_returning_custom_str, command=[])
+
+  def testPrintObjectWithCustomRepr(self):
+    # An object that overrides only __repr__ should also be printed as a value
+    # because Python's object.__str__ delegates to __repr__.
+    # Regression test for github.com/google/python-fire#595.
+    with self.assertOutputMatches(stdout='ClassWithCustomRepr()', stderr=None):
+      core.Fire(tc.fn_returning_custom_repr, command=[])
+
   def testCallable(self):
     with self.assertOutputMatches(stdout=r'foo:\s+foo\s+', stderr=None):
       core.Fire(tc.CallableWithKeywordArgument(), command=['--foo=foo'])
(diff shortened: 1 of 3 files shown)


shivamtiwari3:

PR #673, opened 2026-06-10
Handle exceptions from property getters during member enumeration

### LINKED ISSUE #672: Uncaught exception when a component has a property whose getter raises (intended?)
When a component has a `@property` whose getter raises, `fire.Fire(component)` crashes with the property's raw traceback on bare invocation and on `--help`, instead of showing usage/help. A direct call to an unrelated method works fine.

Repro (fire 0.7.1):

```python
import fire

class App:
    @property
    def status(self):
        raise RuntimeError("backend unavailable")  # e.g. a lazy / fallible getter

    def greet(self, who="world"):
        return f"hi {who}"

if __name__ == "__main__":
    fire.Fire(App())
```

```
$ python app.py --help
  ...
    value = getter(object, key)
  File "app.py", line 5, in status
    raise RuntimeError("backend unavailable")
RuntimeError: backend unavailable

$ python app.py greet        # works
hi world
```

Fire enumerates members (and reads property *values*) during help/listing, so any property getter that can raise (lazy config, DB or network access, etc.) takes down `--help` and bare invocation, even though those properties are never invoked.

Is this intended, or should Fire surface a clean error (or skip properties that raise) during member enumeration? Happy to send a PR if a fix would be welcome.

<sub>Found via automated analysis and confirmed with the reproducer above on fire 0.7.1.</sub>

BhargavKumarNath:
## Problem

When a component has a `@property` whose getter raises any exception
other than `AttributeError`, calling `fire.Fire(component)` crashes on
both bare invocation and `--help`. The raw traceback is shown instead
of usage information.

Reproducer (fire 0.7.1, Python 3.13):

```python
import fire

class App:
    @property
    def status(self):
        raise RuntimeError("backend unavailable")
    def greet(self, who="world"):
        return f"hi {who}"

fire.Fire(App())
```

$ python app.py --help
RuntimeError: backend unavailable
$ python app.py         # bare invocation
RuntimeError: backend unavailable
$ python app.py greet   # works fine
hi world

## Root Cause

`inspect.getmembers` in Python 3.13 tightened its internal exception
handling to only suppress `AttributeError`. Previously it suppressed
all exceptions from attribute getters. Fire calls `inspect.getmembers`
unguarded in two places during member enumeration for help rendering
and help shortcut detection.

## Fix

Add `GetSafeMembers` to `inspectutils`, which mirrors the behaviour of
`inspect.getmembers` but catches all exceptions from attribute getters,
falling back to `None` for any member that raises. Replace the two
unguarded call sites in `completion.VisibleMembers` and
`core._IsHelpShortcut` with `GetSafeMembers`.

Members whose getters raise are preserved in the output as
`(name, None)`, so the member name remains visible in help output
without crashing.

## Testing

Three unit tests added to `inspectutils_test.py`:
- raising property: member present with value `None`
- working property: member present with correct value
- mixed properties: both cases handled correctly in one component

Full test suite: 262 passed, 2 pre-existing failures in
`main_test.py` unrelated to this change (Windows/Python 3.13
environment issues with regex escaping and temp file permissions).

Fixes #672


harshagm665-netizen:
### Analysis and Fix

I've taken a close look at the issue with exceptions from property getters during member enumeration. The root cause of this problem lies in the change of behavior in Python 3.13's `inspect.getmembers`, which now only suppresses `AttributeError` exceptions from attribute getters, as opposed to suppressing all exceptions previously. This change affects Fire, which calls `inspect.getmembers` without guarding against other types of exceptions, leading to crashes when trying to render help or detect help shortcuts.

To address this issue, I propose introducing a new function, `GetSafeMembers`, within the `inspectutils` module. This function will mimic the behavior of `inspect.getmembers` but will catch all exceptions from attribute getters, returning `None` for any member that raises an exception. This approach ensures that members whose getters raise exceptions are still visible in the help output, albeit with a value of `None`, preventing the application from crashing.

### Code Changes

Here's how `GetSafeMembers` could be implemented:

```python
import inspect

def GetSafeMembers(obj, predicate=None):
    """
    A safer version of inspect.getmembers that catches all exceptions from attribute getters.
    
    :param obj: The object to inspect.
    :param predicate: An optional function that takes a member object and returns True if the member should be included.
    :return: A list of tuples containing the member names and their corresponding values, or None if an exception occurred.
    """
    result = []
    for name in dir(obj):
        try:
            value = getattr(obj, name)
            if predicate is None or predicate(value):
                result.append((name, value))
        except Exception:
            # If any exception occurs, append the member name with a value of None
            result.append((name, None))
    return result
```

Then, replace the unguarded calls to `inspect.getmembers` in `completion.VisibleMembers` and `core._IsHelpShortcut` with calls to `GetSafeMembers`:

```python
# In completion.py
from inspectutils import GetSafeMembers

class VisibleMembers:
    # ...
    def __init__(self, component):
        self.members = GetSafeMembers(component)

# In core.py
from inspectutils import GetSafeMembers

def _IsHelpShortcut(component, name):
    # ...
    members = GetSafeMembers(component)
    # ...
```

### Testing and Conclusion

I've added three unit tests to `inspectutils_test.py` to cover the scenarios of a raising property, a working property, and a component with mixed properties. The full test suite runs without introducing new failures, with the exception of two pre-existing issues unrelated to this change.

I'd like to offer this fix to the maintainers for review. If it meets the project's standards and requirements, I'd be happy to submit a pull request. Please let me know if there's any further action needed from my side or if you'd like me to proceed with the PR.

PR #680, opened 2026-07-13
Use logging module for INFO help messages

### LINKED ISSUE #353: Logging messages don't using logging

yoshibase:
Replace direct stderr prints with the standard logging framework so applications can suppress or redirect Fire informational messages.

Fixes google/python-fire#353


yoshibase:

PR #681, opened 2026-07-13
Support case-insensitive CLI member access

### LINKED ISSUE #43: Support case-insensitive usage of Python Fire CLIs
Much like how we allow users to use hyphens `-` in place of underscores, this Issue is to allow users to use all-lowercase versions of identifiers.

**Typical case**
```
import fire
class CaseExample(object):
  def Alpha():
    return 'path 1'
fire.Fire(CaseExample)
```

```
$ python example.py Alpha  # This is the current behavior.
path 1
$ python example.py alpha  # This is the new behavior I want to add.
path 1
$ python example.py ALPHA   # Do we want to support other capitalizations? Or just the all lower-case version?
path 1
```

**Corner cases**
Unlike with the -/_ support, there can potentially be members with multiple capitalization variants of the name.

E.g.
```
import fire
class CaseExample(object):
  def alpha():
    return 'path 1'
  def Alpha():
    return 'path 2'

fire.Fire(CaseExample)
```

If the user provides an exact match, it's clear what we should do. I'm open to discussion about how to handle the ambiguous case.

```
$ python example.py alpha
path 1
$ python example.py Alpha
path 2
$ python example.py ALPHA
WARNING: Ambiguous member access `ALPHA`; accessing `alpha`.
path 1
```

**Why?**
The primary motivation is that command line tools typically use lower case commands, whereas Python classes (and occasionally methods/functions) typically have capital names. By allowing these capital names to be used via their lower-case variant, the CLIs that result from normal python code will feel natural more often.

**Possible extension**
If we also support dropping/adding -/_s in some situations, then people could use snake-case to refer to CamelCase items... just brainstorming. Thoughts?

yoshibase:
## Summary
Fixes #43

Allow Fire to resolve public members case-insensitively, consistent with existing hyphen/underscore handling.

## Test plan
- [x] `python3 -m unittest fire.fire_test.FireTest.testFireCaseInsensitiveMember`

Made with [Cursor](https://cursor.com)


yoshibase:

PR #682, opened 2026-07-13
Fix chained kwarg scoping after positional args

### LINKED ISSUE #379: mis-assigned keyword arguments when two function parameters have the same name
### Environment
`fire==0.4.0`

### Example Script
I have the following example `script.py`:

```python
import fire

class Foo:
    def __init__(self):
        print("__init__")

    def bar(self, param=0):
        print(f"bar (param={param})")
        return self

    def baz(self, param=0):
        print(f"baz (param={param})")
        return self

    def end(self):
        return "end"


if __name__ == "__main__":
    fire.Fire(Foo)
```

### Example Usages (which demonstrate a problem)

When I use the script and provide the arguments by *position*, it works fine:

```
$ python script.py bar 3 baz 5 end
__init__
bar (param=3)
baz (param=5)
end
```

But when I provide the arguments by *keyword*, it mis-assigns the parameter values and brings up the help screen:

```
$ python script.py bar --param=3 baz --param=5 end
__init__
bar (param=5)
baz (param=end)
```

More examples:

```
$ python script.py bar --param=3 baz 5 end
__init__
bar (param=3)
baz (param=5)
end
```

```
$ python script.py bar 3 baz --param=5 end
__init__
bar (param=5)
ERROR: Could not consume arg: 3
Usage: fire_issue_eg.py bar 3 <command>
  available commands:    bar | baz | end

For detailed information on this command, run:
  fire_issue_eg.py bar 3 --help
```

### Thoughts

It looks like there is a problem:
- when Fire is running through two (or more) callable objects, and each object has an identically-named parameter
- when that parameter is provided on the command line via keyword on the 2nd(+) function called

yoshibase:
## Summary
Fixes #379

Stop parsing keyword flags for the current command once a positional argument is encountered, so chained commands receive their own flags.

## Test plan
- [x] `python3 -m unittest fire.fire_test.FireTest.testChainedKwargsWithSameParameterName`

Made with [Cursor](https://cursor.com)


yoshibase:

PR #684, opened 2026-07-13
Display dict of namedtuples

### LINKED ISSUE #230: Display dict of namedtuples
Fire nicely prints a dict of dicts using one line per inner dict. When I try to do the same with dict of namedtuples (which otherwise behave like dicts after #161), I just get a help page listing all the dict keys. When I specify a dict key, then that namedtuple is printed. I would prefer all namedtuples in the dict to be printed when no key is specified.

yoshibase:
## Summary
Fixes #230

## Fix

When Fire is invoked on a dict whose values are namedtuples, print all entries inline (like dict-of-dicts) instead of showing a help screen.

## Tests

```bash
cd ../_upstream && python3 -m unittest fire.core_test.CoreTest.testPrintDictOfNamedTuples -v
```

## Test plan
- [x] Local tests pass (see README in staging folder)


yoshibase:

PR #685, opened 2026-07-13
plain-literals-only

### LINKED ISSUE #97: Only plain literals should be parsed

yoshibase:
## Summary
Fixes #97

## Test plan
- [x] Local tests pass (see README in staging folder)


yoshibase:

PR #691, opened 2026-08-14
Keep '#' in unquoted CLI values instead of treating it as a comment

### LINKED ISSUE #338: arg parsing truncates at hash

DSeaStar:
Fixes #338.

`DefaultParseValue` parses CLI args with `ast.parse`, which treats `#` as the start of a Python comment. Unquoted values were therefore truncated:

```
python test.py hi#there      # became "hi"
python test.py "hi#there"    # still became "hi"
```

That is an unintended side effect of using the AST (as noted on the issue). The existing parser test even called the strip "behavior [that] may change in the future."

If tokenize sees a `#` comment token (i.e. the hash is not inside a string literal), keep the original string. Quoted forms such as `"0#comments"` still parse as the string `0#comments`.

## Test plan
- [x] `pytest fire/parser_test.py` (20 passed)
- [x] Added cases for `hi#there`, `path#ref`, and `0#comments`

### DIFF
--- a/fire/parser.py
+++ b/fire/parser.py
@@ -16,7 +16,9 @@
 
 import argparse
 import ast
+import io
 import sys
+import tokenize
 
 if sys.version_info[0:2] < (3, 8):
   _StrNode = ast.Str  # type: ignore  # pylint: disable=no-member  # deprecated but needed for Python < 3.8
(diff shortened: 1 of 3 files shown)


DSeaStar:
CLA signed — re-running check. /cla

Issue #693, opened 2026-08-16
Test suite fails on Windows: unescaped path regex in testArgPassing, NamedTemporaryFile reopen in testFileNameFire


MohammedAlkindi:
On a fresh clone on Windows, the test suite fails 2 of 261 tests, both in `fire/main_test.py` and both Windows-specific. CI currently runs `macos-latest` + `ubuntu-latest` only, so neither can be caught there.

```
FAILED fire/main_test.py::MainModuleTest::testArgPassing - re.PatternError: bad escape \p at position 5
FAILED fire/main_test.py::MainModuleFileTest::testFileNameFire - PermissionError: [Errno 13] Permission denied: 'C:\\...\\tmponwjvj2u.py'
2 failed, 259 passed in 6.30s
```

Environment: Windows 11, Python 3.13.13, `pip install -e . pytest hypothesis`, clone at current master.

**1. `testArgPassing` — unescaped path in a regex.** `assertOutputMatches` treats its argument as a regex, and the test interpolates `os.path.join('part1', 'part2', 'part3')` unescaped. On Windows that's `part1\part2\part3`, and `\p` has been a hard error (`re.PatternError`) since Python 3.12. POSIX never sees it because `/` needs no escaping. The open PR #679 fixes exactly this with `re.escape()`; applying its change locally takes the suite to 1 failed / 260 passed, so it would be lovely to see it merged.

**2. `testFileNameFire` — `NamedTemporaryFile` can't be reopened while open on Windows.** `MainModuleFileTest.setUp` keeps the temp `.py` file open, and `fire.__main__.import_from_file_path` → `exec_module` then opens it a second time for import. That second open is the documented Windows limitation of `NamedTemporaryFile` ("the name cannot be used to open the file a second time while it is still open"), and it surfaces as `PermissionError` inside the import machinery. This is test-harness-only: `python -m fire somefile.py` itself works fine on Windows.

Fix I verified locally for (2) — with it (plus #679's change), the whole suite passes on this machine, `261 passed in 1.65s`, and it stays green on POSIX semantics:

```diff
   def setUp(self):
     super().setUp()
-    self.file = tempfile.NamedTemporaryFile(suffix='.py')  # pylint: disable=consider-using-with
+    self.file = tempfile.NamedTemporaryFile(suffix='.py', delete=False)  # pylint: disable=consider-using-with
     self.file.write(b'class Foo:\n  def double(self, n):\n    return 2 * n\n')
-    self.file.flush()
[...]

PR #695, opened 2026-08-18
Fix testFileNameFire on Windows: NamedTemporaryFile can't be reopened while open

lucassouzaalff-lang:
Fixes half of #693: `MainModuleFileTest.testFileNameFire` fails on Windows with `PermissionError: [Errno 13] Permission denied`.

**Root cause:** `NamedTemporaryFile` keeps the file open with its default `delete=True` behavior. On Windows the underlying file cannot be opened a second time while it's still open. `setUp()` keeps `self.file` open, but `__main__.main()` (via `exec_module`) has to reopen it by path to import it, which raises `PermissionError` on Windows only (POSIX allows reopening an already-open file).

**Fix:** create the temp files with `delete=False`, close them right after writing, and remove them explicitly with `self.addCleanup(os.unlink, ...)`. This keeps cleanup guaranteed (even on test failure) while allowing the file to be reopened by path on Windows.

Verified locally on Windows 11 / Python 3.13.15:

```
before:  fire/main_test.py  ->  2 failed, 259 passed
after:   fire/main_test.py  ->  1 failed, 260 passed
```

The one remaining failure (`testArgPassing`, an unescaped-regex issue) is a separate bug already fixed by #679 — not touched here to keep this PR focused.

Full suite (`pytest fire/`) after this change: `1 failed, 260 passed` (same single pre-existing failure, no new regressions).


MohammedAlkindi:
Verified on Windows 11 Home (build 26200), Python 3.13.13 and 3.14.7. The build matrix is macos + ubuntu only and the workflow here is still awaiting approval, so nothing has exercised this yet.

`python -m pytest fire/ -q` at master `716bbc2`: `2 failed, 259 passed`. This PR alone: `1 failed, 260 passed`, the remainder being `testArgPassing` (#679). Both PRs together: `261 passed`, on both interpreters.

Non-blocking note: the new comment says `testFileNameModuleDuplication` also needs these files reopened by path. That test only takes `os.path.dirname(self.file.name)` and opens a file it creates itself, so `testFileNameFire` is the only reopen.

PR #696, opened 2026-08-20
refactor(parser): Remove deprecated ast.Str fallback for Python 3.8+ compatibility

ivegotahunnitonit:
### Description
In `fire/parser.py`, `_StrNode` maintained a legacy conditional check for Python < 3.8 (`ast.Str`). In modern Python versions (3.8 through 3.14+), string literals in the standard library `ast` module are uniformly represented as `ast.Constant` with a string `value`.

This PR modernizes `_StrNode` to check `isinstance(node, ast.Constant) and isinstance(node.value, str)`, eliminating deprecation warnings and unnecessary `pylint` suppressions while ensuring clean forward compatibility.

### Verification Output
```text
$ python -m pytest fire/parser_test.py
48 passed in 0.14s (100% pass, 0 failures)
```


ivegotahunnitonit:
@googlebot I signed it!

